Issue #85, opened 2024-05-24
[Mobile-42] Back btn on fully extended sheet on map view 


simon-the-shark:
As a user i want to use back gesture to reset the fully extended sheet to init state

# start state
![Screenshot_20240523_214551.jpg](https://github.com/Solvro/mobile-topwr/assets/28555148/dc828896-637f-4ae0-ad41-8277233005aa)

# action
Back btn or gesture

# desired end state
![Screenshot_20240523_214554.jpg](https://github.com/Solvro/mobile-topwr/assets/28555148/d9561a0c-6e3a-4be1-9031-566a5ee08ee0)

Issue #84, opened 2024-05-24
[Mobile-41] Back btn behaviour on parkings view 


simon-the-shark:
As a user when I use back btn on active parking i want it to go back to full list

# start state
![Screenshot_20240523_214537.jpg](https://github.com/Solvro/mobile-topwr/assets/28555148/cda4b28c-bd30-4423-be2f-115c0da781f9)

# action
Back btn or gesture

# desired end state
![Screenshot_20240523_214544.jpg](https://github.com/Solvro/mobile-topwr/assets/28555148/8845aacc-c6eb-4d6c-a71a-d79263876f80)

simon-the-shark:
Yeah, in buildings maps with activated building - the same story

Issue #65, opened 2024-04-29
[Mobile - 31] Nav bar navigation with custom history (back btn/gesture)


simon-the-shark:
as a user I want to use back button/gesture to return to previous tab
as a user when I go to parking tab from home tab and then use back button to return to home tab

PR #94, opened 2024-06-03
Nested-navigator-based TabBar navigation (Back button improvements)

simon-the-shark:
Implements this feature requests:
- #65 
- #85 
- #84 

I started with simple custom double linked queue in `BottomNavBarController` to keep track of tabs history, but then due to our android workaround I couldn't determine if the `PopScope` blocking the pop is the one that prevents killing the app or is it my custom one in some `MapView` that prevents pop to include it's own logic. I could have still implement these features in some way, but didn't consider it "nice" or "clean" enough.

That's how I ended up implementing fully `Navigator` based navigation in `TabBarView`. Now when we change a tab, what we really do is push new `Route` onto navigator stack. It has proper slide transitions from left or right and when we pop it slides back in reversed direction as well. 

Basically our nested `Navigator` keeps tracks of routes history and handles **any**  navigation in the app now. We can also block any popping e.g. in `MapView` when we want to hide a bottom sheet instead or unselect a marker with simple `PopScope` in this route.

`BottomNavBarController` is now only listening to these `Route`s changes with custom `RouteObserver` and is used to tell the world what the active tab is (e.g. to show proper icon on bottom nav bar). But when we want to make a change, we interact with `Navigator`

Hopefully, it's not an overkill solution :)


lib/config.dart
```diff
@@ -1,10 +1,12 @@
 import 'package:flutter/material.dart';
 import 'package:google_maps_flutter/google_maps_flutter.dart';
 
+import 'features/bottom_nav_bar/nav_bar_config.dart';
 import 'features/iparking_chart/utils/range_hour_points.dart';
 
 abstract class MyAppConfig {
   static const title = "ToPwr";
+  static const initialTab = NavBarEnum.home;
```
mikolaj-jalocha, line 9: I was thinking if we could create separate config for navigator? And place there all navigation's related variables? What do you think?

lib/features/bottom_nav_bar/bottom_nav_bar_controller.dart
```diff
@@ -1,29 +1,24 @@
...
-    return NavBarEnum.home;
+    return MyAppConfig.initialTab;
   }
 
-  void onIndexChanged(int index) {
-    if (index == state.index) {
-      return; // prevents potencial popping of a nested view, when clicked on a current tab
+  void _onTabChanged(NavBarEnum tab) {
+    if (state != tab) {
+      state = tab;
```
mikolaj-jalocha, line 21: Do you think the if statement is needed here? Is the performance significantly better that way? I'm just wondering, because with plain variables, according to my best knowledge, it doesn't.

lib/features/home_view/widgets/departments_section/department_section.dart
```diff
@@ -24,9 +25,7 @@ class DepartmentSection extends ConsumerWidget {
           title: context.localize.departments,
           actionTitle: context.localize.list,
           onClick: () {
-            ref
-                .read(bottomNavBarControllerProvider.notifier)
-                .goTo(NavBarEnum.faculties);
+            ref.read(navigatorProvider).changeTabBar(NavBarEnum.faculties);
```
mikolaj-jalocha, line 28: Maybe we can extract this to a separate variable just like in the `buildings_section.dart` in sake of consistence. 

 ```dart
 static void goToFacultiesTab(WidgetRef ref) =>
      ref.read(navigatorProvider).changeTabBar(NavBarEnum.mapp);
```

File: lib/features/bottom_nav_bar/bottom_nav_bar.dart [hunk 4]
```diff
@@ -1,73 +1,59 @@
...
+          backgroundColor: context.colorTheme.greyLight,
+          showSelectedLabels: false,
+          showUnselectedLabels: false,
+          type: BottomNavigationBarType.fixed,
+          items: NavigationBarItemsList(selectedTab, context),
+        ));
   }
 }
+
+class NavigationBarItemsList extends DelegatingList<BottomNavigationBarItem> {
```
mikolaj-jalocha, line 41: This could be private I guess

lib/features/home_view/widgets/news_section.dart
```diff
@@ -23,9 +24,7 @@ class NewsSection extends ConsumerWidget {
           SubsectionHeader(
               title: context.localize.whats_up,
               onClick: () {
-                ref
-                    .read(bottomNavBarControllerProvider.notifier)
-                    .goTo(NavBarEnum.info);
+                ref.read(navigatorProvider).changeTabBar(NavBarEnum.info);
```
mikolaj-jalocha, line 27: And here as well, maybe it's worth refactoring. Just to be consistent.

lib/features/home_view/widgets/study_circles_section.dart
```diff
@@ -25,9 +26,7 @@ class StudyCirclesSection extends ConsumerWidget {
               title: context.localize.study_circles,
               actionTitle: context.localize.list,
               onClick: () {
-                ref
-                    .read(bottomNavBarControllerProvider.notifier)
-                    .goTo(NavBarEnum.sciCircles);
+                ref.read(navigatorProvider).changeTabBar(NavBarEnum.sciCircles);
```
mikolaj-jalocha, line 29: And here

lib/features/map_view/widgets/bottom_scroll_sheet/bottom_scroll_sheet.dart
```diff
@@ -32,21 +33,24 @@ class BottomScrollSheet<T extends GoogleNavigable> extends ConsumerWidget {
             min(1.0, recomendedSheetHeight / screenHeight);
```
mikolaj-jalocha, line 33: [OFF TOP] Android studio complains that we have a typo here 😝 Should be _recommended_.

lib/features/map_view/widgets/bottom_scroll_sheet/map_view_pop_behaviour.dart
```diff
@@ -0,0 +1,33 @@
...
+  const MapViewPopBehaviour({
+    required this.child,
+    required this.screenHeight,
+    super.key,
+  });
+  final double screenHeight;
+  final Widget child;
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final isAlmostFullyExtended =
```
mikolaj-jalocha, line 18: Is this needed here? I mean, could we extend the bottom sheet to let's say 50% of it's height?

lib/features/map_view/widgets/bottom_scroll_sheet/map_view_pop_behaviour.dart
```diff
@@ -0,0 +1,33 @@
...
+        ref.watch(bottomSheetPixelsProvider) / screenHeight > 0.98;
+
+    final isAnyActive = ref.watch(context.activeMarkerController<T>()) != null;
+
+    return PopScope(
+        canPop: !isAnyActive && !isAlmostFullyExtended,
+        onPopInvoked: (didPop) {
+          if (!didPop) {
+            ref.read(context.mapController<T>().notifier).onMapBgTap(null);
```
mikolaj-jalocha, line 27: I'd consider renaming the `onMapBgTap `to sth more meaningfull like `onMapBackgroundTap`. I took me a while to understand what it exactly does. But maybe it's just me 😝😝

lib/features/navigator/utils/extract_tabbar_arg.dart
```diff
@@ -0,0 +1,13 @@
...
+
+import '../../bottom_nav_bar/nav_bar_config.dart';
+
+extension ExtractTabbarArg on RouteSettings? {
+  NavBarEnum? get selectedTab {
+    final args = this?.arguments;
+    if (args is NavBarEnum) {
+      return args;
+    }
+    return null;
```
mikolaj-jalocha, line 11: I think it's good to change it to sth like this (we stick to that convention in other files and in my opinion it's more readable)
```dart
NavBarEnum? get selectedTab {
    final args = this?.arguments;
    return args is NavBarEnum ? args : null;
  }
```

mikolaj-jalocha (commented):
Hi! Excellent work here!

🔹It was a very good idea to introudce here custom nested navigator since with all the complexities we have here this is the most readable and clean solution although momentarily complex  ut I think it's always that way when dealing with not trivial navigation)

🔹Below I marked some suggestions, observations and questions of mine :-)

lib/features/navigator/navigator/tab_bar_navigator.dart
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    navigatorKey.currentState?.popUntil(
+      (route) => route.settings.name == AppRoutes.root,
+    );
+
+    navigatorKey.currentState?.pushNamed(
+      AppRoutes.root,
+      arguments: newTab,
+    );
```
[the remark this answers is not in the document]
simon-the-shark: oki

lib/features/home_view/widgets/departments_section/department_section.dart
simon-the-shark: sure

lib/features/bottom_nav_bar/bottom_nav_bar.dart
simon-the-shark: well done! very nice catch

lib/features/home_view/widgets/news_section.dart
simon-the-shark: okey. it's defienetly more consistent like this

lib/features/map_view/widgets/bottom_scroll_sheet/bottom_scroll_sheet.dart
simon-the-shark: damn. spelling is kinda hard :P corrected as suggested :)

lib/features/map_view/widgets/bottom_scroll_sheet/map_view_pop_behaviour.dart
simon-the-shark: Yeah, that's actually very good note. it was obviously obvious to me, but it may not be for all people

lib/features/navigator/utils/extract_tabbar_arg.dart
simon-the-shark: matter of preferences I guess, but suggestion applied :)

lib/features/map_view/widgets/bottom_scroll_sheet/map_view_pop_behaviour.dart
simon-the-shark: I added this cause the idea was to block view popping when the bottom sheet is fully extended - and reset the position of the sheet, instead of popping the view. And I've decided to give it a 2% margin of error (Idk for what reasons, just to be safe), but if you think we should remove this failsafe, I'm all for it

lib/config.dart
simon-the-shark: Yeah, actually it gave me the idea to gather all config files to `/config` directory, cause we have bunch of them somewhere in the code. @mikolaj-jalocha what do you think of that?

lib/features/bottom_nav_bar/bottom_nav_bar_controller.dart
simon-the-shark: Actually, as far I know, you are right :P. But I believe I had some problems, that's why I've added this `if` - but maybe it was due to different reasons. I'll defienetly check this out and test again

lib/features/map_view/widgets/bottom_scroll_sheet/map_view_pop_behaviour.dart
mikolaj-jalocha: Thanks for clarification. I thought the same but wanted to be sure. Actually it cost us nothing to have that safety future so in my opinion it is not wrong to have it :-)

lib/config.dart
mikolaj-jalocha: It would be perfect! I like the idea!

lib/features/bottom_nav_bar/bottom_nav_bar_controller.dart
simon-the-shark: So first things, you're right, riverpod checks if the state has changed and rebuilds it only when it did. Here's the riverpod source code
Zrzut ekranu 2024-06-5 o 22 06 02

simon-the-shark: But I obviously knew this. But for a still unknown reasons for me, somehow due to usage our `RouteObserver<PageRoute<dynamic>>` it gives an error - but only if we include this enum in the args of the initial route. If we have null args in the initial route, it doesn't break when we jump to next ones. The problem exists only in first render of the app
Screenshot_1717618165

simon-the-shark: The reason of the error is that it somehow is triggering rebuild of the ui in unfinished state of rendering - which is forbidden due to risk of never ending render (when one render causes re-render before it's ended = never ending single render which trapped in infinite loop) - that's why we see the error. 

But it only breaks during first render and for unknown reasons the `if-check` helps. Which suggests that riverpod's internal check is somehow broken during the initial render. 

Also, for you to know for the future, when you see error like that, the easiest always working fix is to make the change-trigger asonchrynous like that:
Zrzut ekranu 2024-06-5 o 22 12 45
or like that:
Zrzut ekranu 2024-06-5 o 22 13 00
it can be in this place as well:
Zrzut ekranu 2024-06-5 o 22 13 33

but generally speaking, in most cases, it may suggest a wrong usage of something, but the workaround almost always works, but should be used with some moderation. It's better to look for other solution if possible. 

Yeah, but somehow this if also fix this:
Zrzut ekranu 2024-06-5 o 22 15 44

I'll look into this tomorrow more, maybe I'm doing something wrong but this is kinda weird for me.

mikolaj-jalocha: Thank you for explaining all of this. I hope we could find a solution here, maybe the next article is coming 😎 

 For sure at some point in the future I will need to take advantage of advices you gave me about errors like that, thanks for this! 

I'm curious about possible fix but not to block the commit we could leave it that way I guess.

Good job!

lib/features/map_view/widgets/bottom_scroll_sheet/map_view_pop_behaviour.dart
simon-the-shark: cool :)

lib/features/bottom_nav_bar/bottom_nav_bar_controller.dart
simon-the-shark: Yeah, maybe it is😎 But rn the university is killing me, I'm already half dead. When doing quick tests, it seems like the riverpod doesn't call updateShouldNotify in this first build? idk it's werid
Zrzut ekranu 2024-06-5 o 22 45 13

simon-the-shark: I'll leave it with `if` and we'll look into that after I pass this term (hopefully)

simon-the-shark:
@mikolaj-jalocha  I believe I've applied the requirements. Very nice review :) - you're becoming a pro in this already. Thanks for all and let me know if I can merge :)

mikolaj-jalocha:
> @mikolaj-jalocha I believe I've applied the requirements. Very nice review :) - you're becoming a pro in this already. Thanks for all and let me know if I can merge :)

LGTM!

Issue #54, opened 2024-04-25
[mobile- 23] About us page


AleksaZakrzewska:
Include a link to Solvro's website and social media on the “about us” page according to the design

![image](https://github.com/Solvro/mobile-topwr/assets/91432120/dbca1304-cd58-4287-b419-d9b61af2947c)

PR #102, opened 2024-06-15
Mobile 23

kamilfedio:
#54 

- created an additional tab (guide) to be able to work on the 'about us' tab - only temporary
- created 'about us' tab with according to the design
- added aboutus theme 
- buttons and icons lead to matching web pages

I am not sure about adding every padding, heigh etc. to config file so i added only repetetive values.
probably some used widgets should be moved to lib/widgets but I'd rather make sure about that.


lib/features/guide/widgets/about_us/models/link_data.dart
```diff
@@ -0,0 +1,32 @@
...
+  String iconUrl;
+  String? url;
+
+  static const _icons = {
+    "facebook": "assets/icons/ic_fb.png",
+    "instagram": "assets/icons/ic_insta.png",
+    "linkedin": "assets/icons/ic_linkedin.png",
+    "github": "assets/icons/ic_github.png",
+    "mailto:": "assets/icons/ic_email.png",
+    "pwr.edu.pl": "assets/icons/ic_web.png",
```
simon-the-shark, line 15: it this `pwr.edu.pl` necessary? I believe when nothing is found, we have a defaultIconUrl which is basically this web icon

lib/features/guide/widgets/about_us/models/link_data.dart
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  LinkData({
```
simon-the-shark, line 18: so this whole file is basically repeated version of `lib\features\study_circle_details\smth...` 

it's probably good idea to make it a global util/model and use it anywhere in the app we need it

#### proposition
my propo would maybe be:
1. `lib\config\url_icons.dart`:
```dart
abstract class UrlIconsConfig {
 static const icons = {
     "facebook": "assets/icons/ic_fb.png",
    "instagram": "assets/icons/ic_insta.png",
    "linkedin": "assets/icons/ic_linkedin.png",
    "github": "assets/icons/ic_github.png",
    "mailto:": "assets/icons/ic_email.png",
  };
}
```

4.  combine `ContactSectionData` and `LinkData`  into  `lib/utils/urls_icons.dart`:

```dart
@immutable
class UrlIconsUtilModel {
  final String iconUrl;
  final String? text;
  final String? url;

  UrlIconsUtilModel({
    String? text,
    this.url,
  })  : iconUrl = _determineIcon(url),
        text = text ?? url;

  static String _determineIcon(String? url) {
    return url != null
        ? UrlIconsConfig.icons.entries
                .firstWhereOrNull(
                  (e) => url.contains(e.key),
                )
                ?.value ??
            DetailsScreenConfig.defaultIconUrl
        : DetailsScreenConfig.defaultIconUrl;
  }
}
```

5. delete both your `LinkData` and original `ContactSectionData` and use our new util model everywhere

lib/features/guide/widgets/about_us/repository/about_us_repository.dart
```diff
@@ -0,0 +1,36 @@
...
+    if (event == null) return null;
+    return AboutUsDetails(
+      aboutUs: event.AboutUs,
+      aboutUsTeam: event.AboutUs_Team,
+    );
+  });
+}
+
+
+class AboutUsDetails {
```
simon-the-shark, line 31: 1. if you want custom model, it shoudl probably go `/model` folder
2. it's best to use `freezed` package for such models
3.  Do you really need this here? it basically works the same as the generated one

lib/features/guide/widgets/about_us/models/member_data.dart
```diff
@@ -0,0 +1,20 @@
+import 'link_data.dart';
+
+class MemberData {
```
simon-the-shark, line 3: so `freezed` is best practice for models, but here it may be impossible/hard to implement (due to a bit weird LinkData class), so that's not a big problem 

but still at least do the following:
1. add `final` to all attributes
2. annotate `immutable` to the class:
```dart
@immutable
class MemberData {
  final List<String> socialLinks;
  final String? name;
  final String? subtitle;
  final List<LinkData> links;
  final String? imageUrl;

//...
```

that's the minimum, for fully proper model you would need to override `==` operator and `hashCode`, but as long it's not fully needed, you can skip it. `freezed` does that for us, but here it may be hard to do, cause we don't want to iterate over the all entries of the map with icons. So just do the 2 things noted above

lib/features/guide/widgets/about_us/widgets/desription_section.dart
```diff
@@ -0,0 +1,47 @@
...
+              if (element.localName == 'a') {
+                return {
+                  'color':
+                      '#${context.colorTheme.orangePomegranade.value.toRadixString(16).padLeft(8, '0').substring(2)}',
+                  'text-decoration': 'none',
+                };
+              }
+              return null;
+            },
+            //todo
```
simon-the-shark, line 38: what's this `//todo` commend doing here :P?

lib/features/guide/widgets/about_us/widgets/desription_section.dart
```diff
@@ -0,0 +1,47 @@
...
+          borderRadius: BorderRadius.circular(AboutUsConfig.borderRadius),
+          color: context.colorTheme.greyLight,
+        ),
+        child: Padding(
+          padding: const EdgeInsets.all(AboutUsConfig.defaultPadding),
+          child: HtmlWidget(
+            text,
+            textStyle: context.aboutUsTheme.body,
+            customStylesBuilder: (element) {
+              if (element.localName == 'a') {
```
simon-the-shark, line 29: I don't like this `'a'` magic string

but I see it's hard to avoid this with this `HtmlWidget` :/

the best I can come up with is maybe smth similar to this in separate file:
```dart
import 'package:flutter/material.dart';
import 'package:html/dom.dart' as html;

import '../../../../../theme/app_theme.dart';
import '../../../../../theme/hex_color.dart';

extension ToHtlColorString on HexColor {
  String get htmlFormat =>
      '#${value.toRadixString(16).padLeft(8, '0').substring(2)}';
}

extension IsLinkTag on html.Element {
  bool get isLink {
    return localName == "a";
  }
}

extension CustomHtmlStyles on BuildContext {
  Map<String, String>? customStylesBuilder(html.Element element) {
    if (element.isLink) {
      return {
        'color': colorTheme.orangePomegranade.htmlFormat,
        'text-decoration': 'none',
      };
    }
    return null;
  }
}
```

lib/features/guide/widgets/about_us/about_us_tab.dart
```diff
@@ -0,0 +1,94 @@
...
+            ),
+            SliverList(
+              delegate: SliverChildListDelegate(
+                [
+                  SectionHeader(text: context.localize.about_us),
+                  DescriptionSection(text: value?.aboutUs?.description ?? ''),
+                  SectionHeader(text: context.localize.meet_out_team),
+                  TeamSection(
+                    members: value?.aboutUsTeam?.whereNonNull
+                            .toList()
```
simon-the-shark, line 60: actually this `toList` shouldn't be here, as we want one single `toList` at the end of this chain of `iterable` operations like `whereNonNull`, `map`, `where`, etc. Cause all of them are lazily evaluated and `toList` actually makes the iteration and convert all this condition to list in a single operation

if you know what i mean. We use basically use `toList` as a final step, and if we only want to make a simple single iteration, we may not even need it at all

simon-the-shark: also my `whereNonNull` can be evoked on nullable type as well and has built-in empty list default.

so you can change `aboutUsTeam?.whereNonNull` to `aboutUsTeam.whereNonNull` and you can delete ` ?? List.empty()` then

lib/features/guide/widgets/about_us/about_us_tab.dart
```diff
@@ -0,0 +1,94 @@
...
+              ),
+            ),
+            SliverList(
+              delegate: SliverChildListDelegate(
+                [
+                  SectionHeader(text: context.localize.about_us),
+                  DescriptionSection(text: value?.aboutUs?.description ?? ''),
+                  SectionHeader(text: context.localize.meet_out_team),
+                  TeamSection(
+                    members: value?.aboutUsTeam?.whereNonNull
```
simon-the-shark, line 59: all this long transformations belong more to `model`/`domain` layer than `ui`, so you can add them as methods to your custom model or as methods in `extension` on a generated one

what i mean:

```dart
class AboutUsDetails {
  final AboutUs? aboutUs;
  final List<AboutUsTeam>? aboutUsTeam;

  AboutUsDetails({this.aboutUs, this.aboutUsTeam});

  //add this
  List<MemberData> getMemberData(List<AboutUsTeam>? aboutUsTeam) {
    return aboutUsTeam.whereNonNull
        .map(
          (e) => MemberData(
            socialLinks: e.socialLinks.whereNonNull.map((e) => e.url).toList(),
            name: e.name,
            subtitle: e.subtitle,
            imageUrl: e.photo?.filename_disk,
          ),
        )
        .toList();
  }

// ...
```

lib/features/guide/widgets/about_us/about_us_tab.dart
```diff
@@ -0,0 +1,94 @@
...
+                                  subtitle: e.subtitle,
+                                  imageUrl: e.photo?.filename_disk ?? ''),
+                            )
+                            .toList() ??
+                        List.empty(),
+                  ),
+                  SectionHeader(text: context.localize.follow_solvro),
+                  LinksSection(
+                    links: value?.aboutUs?.solvroSocialLinks.whereNonNull
+                            .map(
```
simon-the-shark, line 78: this transformation should go to model too

simon-the-shark (commented):
@kamilfedio good job done, I've left bunch of note in few places. Always feel free to ask questions or discuss if you think i'm wrong. 

 🦈

lib/features/guide/guide_view_template.dart
```diff
@@ -0,0 +1,50 @@
...
+        title: const Text('O nas'),
+        backgroundColor: context.colorTheme.greyLight,
+      ),
+      body: Padding(
+        padding: const EdgeInsets.all(20.0),
+        child: Column(
+          children: [
+            GestureDetector(
+              onTap: () {
+                Navigator.push(
```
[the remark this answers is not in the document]
kamilfedio: I did it temporarily for the purpose of my widget to be able to redirect, and then remove it, but I will correct it

lib/features/guide/widgets/about_us/widgets/desription_section.dart
kamilfedio: I wasn't sure if this is how lanuch should be handled so I marked it as //todo :d

lib/features/guide/widgets/about_us/about_us_tab.dart
kamilfedio: when i tried to remove `?` i got an error about nullable list

kamilfedio:
@simon-the-shark I think I have already corrected all the mistakes, but could you check if I understood some of them correctly?

lib/features/guide/widgets/about_us/about_us_tab.dart
simon-the-shark: I'm almost sure `whereNonNull` is an extension on both nullable and nonnullable lists

lib/features/guide/widgets/about_us/models/about_us_details.dart
```diff
@@ -0,0 +1,34 @@
...
+import '../repository/about_us_repository.dart';
+import 'member_data.dart';
+
+class AboutUsDetails {
+  final AboutUs? aboutUs;
+  final List<AboutUsTeam>? aboutUsTeam;
+
+  AboutUsDetails({this.aboutUs, this.aboutUsTeam});
+
+  List<MemberData> getMemberData(List<AboutUsTeam>? aboutUsTeam) {
```
simon-the-shark, line 13: why do you accept `List<AboutUsTeam>? aboutUsTeam` here as argument when you basically have this as class's field?

lib/features/guide/widgets/about_us/about_us_tab.dart
```diff
@@ -56,30 +56,11 @@ class _AboutUsView extends ConsumerWidget {
...
-                                          .map((e) => e.url)
-                                          .toList() ??
-                                      List.empty(),
-                                  name: e.name,
-                                  subtitle: e.subtitle,
-                                  imageUrl: e.photo?.filename_disk ?? ''),
-                            )
-                            .toList() ??
-                        List.empty(),
+                    members: AboutUsDetails().getMemberData(value?.aboutUsTeam),
```
simon-the-shark, line 59: When you remove the useless argument from this method you should be able to call 
```dart
value?.aboutUsTeam.getMemberData()
```

or am i wrong? haven't tested it myself

lib/features/guide/widgets/about_us/models/about_us_details.dart
```diff
@@ -0,0 +1,34 @@
...
+            imageUrl: e.photo?.filename_disk?.directusUrl,
+            socialLinks:
+                e.socialLinks?.whereNonNull.map((e) => e.url).toList() ?? [],
+            subtitle: e.subtitle,
+          );
+        }).toList() ??
+        [];
+  }
+
+  List<UrlIconsUtilModel> getSocialIcons(AboutUs? aboutUs) {
```
simon-the-shark, line 26: same here?

simon-the-shark (commented):
one last note about two methods and you can merge

lib/features/guide/widgets/about_us/models/about_us_details.dart
kamilfedio: I tried but only to do it is something like this
`TeamSection(
                    members: value?.getMemberData() ?? [],
                  ),`
and function:
`List<MemberData> getMemberData() {
    return aboutUsTeam.whereNonNull.map((e) {
      return MemberData(
        name: e.name,
        imageUrl: e.photo?.filename_disk?.directusUrl,
        socialLinks:
            e.socialLinks.whereNonNull.map((e) => e.url).toList(),
        subtitle: e.subtitle,
      );
    }).toList();
  }`
  
  Is this what you were thinking about?

simon-the-shark (approved):

Issue #101, opened 2024-06-10
[Mobile -53] Add Youtube icon?


simon-the-shark:
Maybe let's add youtube icon as well:
<img width="149" alt="Zrzut ekranu 2024-06-10 o 17 46 58" src="https://github.com/Solvro/mobile-topwr/assets/28555148/217eec08-2da2-4509-9e7c-146e5dfe360c">

mikolaj-jalocha:
Do we have somewhere youtube icon? 
@simon-the-shark @gry-mar

simon-the-shark:
Yeah, good point, I haven't seen any on Figma yet.

simon-the-shark:
But basically you just gotta take some random youtube svg and paint it pomegranade - even I can do it for you in a moment. Unless we're using some specific icon package or smth @gry-mar ?

simon-the-shark:
Also, we do have a github icon, so you can add it there as well, and we'll have some global utils for detecting url's icons this way

simon-the-shark:
I guess smth like this will do?

<img width="387" alt="Zrzut ekranu 2024-06-15 o 00 14 39" src="https://github.com/Solvro/mobile-topwr/assets/28555148/be7c2b1f-f1b4-48a9-928f-6461f7d7735c">

simon-the-shark:
![yt](https://github.com/Solvro/mobile-topwr/assets/28555148/389c640c-3619-4ffa-a870-bec076447ccc)

PR #103, opened 2024-06-19
Add youtube and github icon [MOBILE 53]

mikolaj-jalocha:
[MOBILE 53](https://github.com/Solvro/mobile-topwr/issues/101)

- add youtube and github icon
- refactor function responsible for determining proper icon


lib/utils/determine_icon.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:collection/collection.dart';
+
+import '../config/ui_config.dart';
+
+extension IconDeterminer on String? {
+  String determineIcon() {
+    const icons = {
```
simon-the-shark, line 7: Maybe this map should go to separate config file in `/config` dir?

simon-the-shark (commented):
One note :P
Goodnight :)

lib/utils/determine_icon.dart
mikolaj-jalocha: Good Morning :-) 
Good point, why not.

Issue #97, opened 2024-06-04
[Mobile-51] More intuitive buildings search


simon-the-shark:
Two small features:
1. When I type "serowiec", "sks" or bibliotech I would like to find those buildings as well (There already is a proper field on backend)
2. When I type "c13" I want to find "c-13" building too

PR #107, opened 2024-06-22
feature: add new filter to bulding search

kamilfedio:
#97 
I added the filters mentioned and improve graphql&model of bulidings

### DIFF
--- a/lib/features/buildings_map/controllers.dart
+++ b/lib/features/buildings_map/controllers.dart
@@ -7,6 +7,7 @@ import '../map_view/controllers/active_map_marker_cntrl.dart';
 import '../map_view/controllers/controllers_set.dart';
 import '../map_view/controllers/map_controller.dart';
 import '../map_view/controllers/map_data_controller.dart';
+import 'utils.dart';
 
 part "controllers.g.dart";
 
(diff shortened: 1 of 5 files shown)


lib/features/buildings_map/controllers.dart
```diff
@@ -34,7 +34,11 @@ class BuildingsListViewController extends _$BuildingsListViewController
   @override
   bool filterMethod(BuildingModel item, String filterStr) {
     return item.name.toLowerCase().contains(filterStr) ||
-        (item.addres?.toLowerCase().contains(filterStr) ?? false);
+        (item.addres?.toLowerCase().contains(filterStr) ?? false) ||
+        (item.naturalName?.toLowerCase().contains(filterStr) ?? false) ||
+        (!filterStr.contains('-')
+            ? item.name.toLowerCase().replaceAll('-', '').contains(filterStr)
```
simon-the-shark, line 40: it stopped being easily readable at this point and I would strongly recomend some refactor to divide this into smaller logically separated pieces.

so you can do this in many ways, but my personal proposition would be:

in map view config file
```dart
abstract class BuildingSearchConfig {
  static const buildingCodeSeparator = "-";
}
```

then create util file, example location: lib\features\buildings_map\utils.dart
```dart
extension ContainsCaseUnsensitive on String? {
  bool containsUnsensitive(String str) {
    if (this == null) return false;
    return this!.toLowerCase().contains(str.toLowerCase());
  }

bool containsBuildingCode(String buildingCode) {
    if (buildingCode.contains(BuildingSearchConfig.buildingCodeSeparator)) {
      return containsUnsensitive(buildingCode);
    }
    return this
            ?.replaceAll(BuildingSearchConfig.buildingCodeSeparator, "")
            .containsUnsensitive(buildingCode) ??
        false;
  }
}
```
and the usage would be:
```dart
  @override
  bool filterMethod(BuildingModel item, String filterStr) {
    return item.name.containsBuildingCode(filterStr) ||
        item.addres.containsUnsensitive(filterStr) ||
        item.naturalName.containsUnsensitive(filterStr);
  }
```

it's just a proposition, you can refactor it in some other way you find suitable. Also I hope I haven't broken the logic :)

simon-the-shark (commented):
good work. I have one proposition of a simple refactor.

kamilfedio:
@simon-the-shark Okay, I used your idea

simon-the-shark:
You can merge :)

PR #111, opened 2024-07-21
Refactor/assets icons fonts

simon-the-shark:
- changed icons from `png` to `svg`
- changed variable font to static values (previous had problems. I just looks like flutter works better with static fonts definitions)
- implemented `flutter_gen` for generating assets path (no more unsafe strings)


mikolaj-jalocha (approved):
LGTM! Icons refactor was a thing that we should fix and standarize long time ago. Especially automatically generated assets path. I will create task to refactor existing parts of code in CatchQR to use there. 

One suggestion from me, what about quick mention of new way of assets generation in readme? So that no one is confused. Probably later on we could delete it.

mikolaj-jalocha:
> LGTM! Icons refactor was a thing that we should fix and standarize long time ago. Especially automatically generated assets path. I will create task to refactor existing parts of code in CatchQR to use there.
> 
> One suggestion from me, what about quick mention of new way of assets generation in readme? So that no one is confused. Probably later on we could delete it.

I wonder if we could create wiki section for our project? Some repos on GitHub use it, especially large ones

simon-the-shark:
Sure thing

PR #118, opened 2024-08-01
Refactor/ navigation remake attempt

simon-the-shark:
## Preview

https://github.com/user-attachments/assets/5886056e-f511-4dda-a42c-eee8b66cc46d

## PR Desc
Small remake of the navigation (yet again:))

1. small animations/transitions changes, but this can be easily changeable 
- tab bar has slide + fade (prev only slide)
- details have slide from bottom
- both have 200ms duration
2. Refactor to type-safe generated routes with `auto_route` and other small refactoring.
3. Now we preserve state of every route - hopefully flutter will handle big and fat statefull navigation stack (my device didn't have any issues in tests) - but it's something to have in mind the future (I have ideas for potencial fixes, but they introduce some complexity)
4. Behaviour/principials changes:
- we pop detail views only when we change tab using bottom navbar. ONLY. This means no popping in other cases like "lista >' button in some other detail view
- we only change tabs when explicitly asked. So when we enter detail view from homescreen directly (or from other detail view) - we do not change tab. We change tabs only when using bottom navbar and "lista >" ref buttons etc.

tell me what you think


simon-the-shark:
#130 reverses detail views transition builder to default one

lib/config/transitions.dart
```diff
@@ -0,0 +1,11 @@
...
+
+// TODO(simon-the-shark): adjust this settings if desired
+abstract class TransitionsConfig {
+  static const durationInMiliseconds = 200;
+  static const detailDurationInMiliseconds = 200;
+  static const slideLeftBuilder = TransitionsBuilders.slideLeftWithFade;
+  static const slideRightBuilder = TransitionsBuilders.slideRightWithFade;
+  static const fallbackBuilder = TransitionsBuilders.fadeIn;
+  static const detailViewBuilder = TransitionsBuilders.slideBottom;
+}
```
mikolaj-jalocha, line 11: Actually very good that they're parametrized in one section. Imo  UX's nice.

lib/features/home_view/widgets/buildings_section/buildings_section.dart
```diff
@@ -73,7 +70,7 @@ class _DataListBuildingsTiles extends ConsumerWidget {
             buildingName: mapItem.name,
             imageUrl: mapItem.cover?.filename_disk?.directusUrl,
             onTap: () async {
-              await BuildingsSection.goToMapTab(ref);
+              unawaited(ref.navigateBuildings());
```
mikolaj-jalocha, line 73: `unawaited `here's very very nice since we don't use the Future

lib/features/navigator/utils/navigation_commands.dart
```diff
@@ -0,0 +1,47 @@
+import "package:flutter_riverpod/flutter_riverpod.dart";
```
mikolaj-jalocha, line 1: Initially I was wondering "from where the `ref.navigateHome()` come from" extremely useful extension here.

mikolaj-jalocha (approved):
Excellent work here. I'm not a navigation Sensei here but even for me (beginner in complex Flutter's concepts, eg. navigation 🙃) everything looks nice and cool. Personally I didn't notice sth strange. 

 I checked the` `[auto_route](https://pub.dev/packages/auto_route) lib, excellent in docs and easy-to-use. Wejust need to add annotation and configure route in controller...  Good choice I guess. WDYT about updating readme? Even with some refrence to docs, like with internalization module. 

In my opinion the nav stack in typical scenario won't be extremely large. So this should not be the problem. Even Android Studio emulator did well.

 We could always think of adding feature for tracking app performance and watch the behaviour on prod. Once I read sth about it [here](https://uxcam.com/blog/flutter-performance-monitoring/). Wonder what u think ;)

lib/features/home_view/widgets/buildings_section/buildings_section.dart
simon-the-shark: Linter screams if you don't do it :) now you have to either await the future or unwait

lib/features/navigator/utils/navigation_commands.dart
simon-the-shark: Yeah, I've been playing around with different approaches and this in one place was usefull when making any changes

simon-the-shark:
Thanks for checking out the changes. Readme link is a good idea too. And about the navigation stack, let's hope for now that flutter handle its well on its own. And yeah, performance and crashlytics are probably a must have before production

PR #121, opened 2024-08-03
Refactor/ unify codebase naming

simon-the-shark:
# Re-pr of #119 

Unified names refactor with simple guidelines and note in readme. You can take a quick look, but I wouldn't spend too much time looking into this, this should be semantically the same, just unified names. We can also obviously discuss some naming choices. Tell me what you think


mikolaj-jalocha (approved):
LGTM!

PR #122, opened 2024-08-03
Fix/ eliminate jumping scroll in home view

simon-the-shark:
Don't be scared with stacked PRs :) this one is very simple. Just make reviews of every one of them and I'll merge them in the end together.

# FIx of #104 
- Actually it is observable on android too, but android has a bit different scroll mechanics, so in ios it was more obvious
- very simple fix, just as I anticipated, one section had different hight while loading than in loaded state and with `ListView.builder` it was rebuilding just in that specific moment and trigger the rebuild and weird scroll mechanics.
- Fixed with simple calculations and sized box.

# Previews
## Issue (before changes)

https://github.com/user-attachments/assets/3f5303c6-4758-4e30-8a1f-e07f44191e46

## Results

https://github.com/user-attachments/assets/3056ea0c-c801-4b1a-aba4-d172eed9c8f5

# Changes. 
## News section. 
this was main problem.  
### Loaded state (no changes). 

Zrzut ekranu 2024-08-3 o 14 58 14

### Loading state BEFORE. 

Zrzut ekranu 2024-08-3 o 14 59 21

### Loading state AFTER. 

Zrzut ekranu 2024-08-3 o 15 22 16

# Alternative
Alternative solution is to preserve `newsRepository` provider in the memory, either with `keepAlive: true` (don't like the idea) or just subscribing to it in the lower section of the home view (feel very hacky). But if you think current loading looks significatly worse, we can solve this this way too.

### DIFF
--- a/lib/config/map_view_config.dart
+++ b/lib/config/map_view_config.dart
@@ -36,8 +36,10 @@ abstract class MapWidgetConfig {
   );
   static const mapType = MapType.normal;
 
-  static const mapMarkerOriginWidth = 28;
-  static const activeMapMarkerOriginWidth = 40;
+  static const mapMarkerWidth = 22.0;
+  static const mapMarkerHeight = 34.0;
+  static const activeMapMarkerWidth = 30.4;
+  static const activeMapMarkerHeight = 46.4;
 }
 
 abstract class BuildingSearchConfig {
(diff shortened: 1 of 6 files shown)


mikolaj-jalocha (approved):
WDYT about changing loading widget to contain slightly larger body (the gray area)? Just to fill the free space. Even with plain gray block or with more lines?  Just a proposition of mine. 

Same as you, I feel very uneasy about handling the reference down...

simon-the-shark:
Very good idea, bigger body might improve its look a bit. I'll try something in a second. But also in most scenarios this won't be visible for too long if even visible at all

Issue #95, opened 2024-06-04
[Mobile - 49] Release build: small map markers (not always)


simon-the-shark:

simon-the-shark:
How it is 
![Screenshot_20240604_142857.jpg](https://github.com/Solvro/mobile-topwr/assets/28555148/611e0b44-c24f-4806-9d35-53a9ab8dd942)

How it is supposed to be
![Screenshot_20240604_091257.jpg](https://github.com/Solvro/mobile-topwr/assets/28555148/ee948e24-9bce-4386-8f6e-3e8a1f03b829)

On debug build it always works and on release sometimes :P

simon-the-shark:
Now they seem to be extra 4x bigger than normal 😭

PR #130, opened 2024-08-03
Fix/ normalize markers size

simon-the-shark:
# Fix of #95 

TODO: Test on multiple devices

# How it looks on my device now
image
image

# Update

### DIFF
--- a/lib/config/map_view_config.dart
+++ b/lib/config/map_view_config.dart
@@ -36,8 +36,10 @@ abstract class MapWidgetConfig {
   );
   static const mapType = MapType.normal;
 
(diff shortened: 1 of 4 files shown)


mikolaj-jalocha (approved):
LGTM.  I did test on Android Device emulator (Pixel 7). 

One thing'd consider to change is the active marker's size. Imo slightly too large. But that's just my suggestion

simon-the-shark:
I've tried to imitate 1:1 from figma, but yeah we can make the active one a bit smaller.

simon-the-shark:
I've set it to 80% of the size it was

PR #146, opened 2024-08-07
fix: add extension to change null address to new value

kamilfedio:


lib/features/buildings_view/utils.dart
```diff
@@ -16,3 +17,13 @@ extension ContainsCaseUnsensitive on String? {
         false;
   }
 }
+
+extension ChangeNullAdress on String? {
+  String? changeNull() {
+    if (this == null) {
+      return AppLocalizationsPl().no_address;
```
simon-the-shark, line 24: Tbh, hardcoding here `AppLocalizationsPl` kills the purpose of localization. You should always get the locale from `BuildContext` to make it work as intended: 
```dart
extension ChangeNullAdressX on BuildContext {
  String? changeNull(String? address) {
    return address ?? localize.no_address;
  }
}
```

simon-the-shark (commented):
one note :)

lib/features/buildings_view/utils.dart
kamilfedio: ok done

simon-the-shark (approved):

PR #150, opened 2024-08-09
Feat/open buildings from contact

simon-the-shark:
I've been fixing some issues that occured during filling the info in the backend and done the opening links as well (haven't seen your assignemnet @mikolaj-jalocha ) Upsik

- done this map links opening
- changed `plac` to `pl.` to be consistent with buildings
- fixed tab bar transition direction in some cases
- upgraded flutter version to `3.24.0`

### DIFF
--- a/lib/features/about_us_view/widgets/contact_section.dart
+++ b/lib/features/about_us_view/widgets/contact_section.dart
@@ -1,4 +1,5 @@
 import "package:flutter/material.dart";
+import "package:flutter_riverpod/flutter_riverpod.dart";
 import "package:flutter_svg/svg.dart";
 
 import "../../../config/ui_config.dart";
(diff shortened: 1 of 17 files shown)


lib/features/about_us_view/widgets/contact_section.dart
```diff
@@ -27,17 +28,17 @@ class ContactSection extends StatelessWidget {
...
   final String url;
   final String icon;
 
   @override
-  Widget build(BuildContext context) {
+  Widget build(BuildContext context, WidgetRef ref) {
     return GestureDetector(
       onTap: () async {
-        await LaunchUrlUtil.launch(url);
+        await ref.launch(url);
```
mikolaj-jalocha, line 41: Can we use here `unawaited`?

lib/utils/launch_url_util.dart
```diff
@@ -1,11 +1,36 @@
...
 import "package:url_launcher/url_launcher.dart";
 
-class LaunchUrlUtil {
-  static Future<bool> launch(String uriStr) async {
+import "../features/buildings_view/repository/buildings_repository.dart";
+import "../features/navigator/utils/navigation_commands.dart";
+import "../utils/where_non_null_iterable.dart";
+
+extension LaunchUrlUtilX on WidgetRef? {
+  Future<bool> launch(String uriStr) async {
```
mikolaj-jalocha, line 10: I like the change here. Universal in all places rn.

mikolaj-jalocha (approved):
Looks good to me :-) i especially like the  new way of launching` url utils` with help of `ref`. 

I left only one note regarding minor change.

lib/features/about_us_view/widgets/contact_section.dart
simon-the-shark: Sure, but also there's nothing after it, so 
```dart
onTap: () async => ref.launch(url),
```
should have similar effect I believe

PR #152, opened 2024-08-10
Fix/129 fix eliminate parkings flickering

mikolaj-jalocha:
I've tested it on Android device (the bug was there before as well). Everything looks fine. Hopefully new state is being updated.


lib/features/parkings_view/repository/parkings_repository.dart
```diff
@@ -13,7 +13,7 @@ extension RefIntervalRefreshX on Ref {
   void setRefresh(Duration interval) {
     final timer = Timer.periodic(
       interval,
-      (t) => invalidateSelf(),
+      (t) => invalidate(parkingsRepositoryProvider),
```
simon-the-shark, line 16: Unfortunetly this throws :( 
```console
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: 'package:riverpod/src/framework/element.dart': Failed assertion: line 614 pos 11: 'listenable._origin != origin': A provider cannot depend on itself
```
basically `invalidateSelf()` and `invalidate(someProvider)` works the same with the difference the second one seems to throw when use inside the same provider (I didn't knew that actually, but make sense)

simon-the-shark (commented):
I'll try to do this too when I have a moment. I have few ideas but we'll see how they'll do

simon-the-shark:
@mikolaj-jalocha I've been playing around with it for some time 😭 But I have a fix. Suprisingly it wasn't so much about the riverpod but more with the UI itself and the `switch` pattern matching logic. Tell me what think. Let's hope everything works as expected now

https://github.com/user-attachments/assets/413a8658-cc7c-43f9-bf7c-b199cc393243

simon-the-shark:
@mikolaj-jalocha when you have a moment you can make a review here, cause i can request review from its author

mikolaj-jalocha:
@simon-the-shark  thanks for feedback. I didn't see exception before (I should look on logs..).

I will do review ASAP (tomorrow)

lib/features/parkings_view/api_client/iparking_client.dart
```diff
@@ -6,7 +6,7 @@ import "../../../config/api_base_config.dart";
 part "iparking_client.g.dart";
 
 abstract class ParkingsConfig {
-  static const parkingsRefreshInterval = Duration(minutes: 1);
+  static const parkingsRefreshInterval = Duration(seconds: 30);
```
mikolaj-jalocha, line 9: Is the period here ok?

lib/features/map_view/widgets/map_widget.dart
```diff
@@ -38,30 +22,40 @@ class _MapWidget<T extends GoogleNavigable> extends ConsumerWidget {
...
-            top: MediaQuery.paddingOf(context).top,
-            bottom: ref.watch(bottomSheetPixelsProvider),
-          ),
+      AsyncValue(:final value) => Consumer(
+          builder: (BuildContext context, WidgetRef ref, Widget? child) {
+            final asyncItems = ref
+                .watch(context.mapSourceRepository<T>())
+                .value
+                .whereNonNull
+                .toList();
```
mikolaj-jalocha, line 35: I wonder if we can use here immutable list? (`IList`)

btw the way `Consumer `here's introduced improved visibility a lot imo.

mikolaj-jalocha: Btw, is this a change that reduced flickering? Moving state refreshment inside `AsyncValue`?

lib/features/map_view/widgets/bottom_scroll_sheet/data_list.dart
```diff
@@ -16,19 +17,21 @@ class DataSliverList<T extends GoogleNavigable> extends ConsumerWidget {
...
-      AsyncLoading() => const DataListLoading(),
+      // misz-masz here fixed flickering
+      // we probably should consider similar refactoring to all switches, but won't be visible in our other use-cases
+      AsyncValue(:final IList<T?> value) =>
+        _DataSliverList<T>(value.whereNonNull.toIList()),
       AsyncError(:final error) =>
         SliverToBoxAdapter(child: MyErrorWidget(error)),
-      AsyncValue(:final value) =>
-        _DataSliverList<T>(value.whereNonNull.toList())
+      _ => const DataListLoading(),
```
mikolaj-jalocha, line 26: What happened to `AsyncLoading`?

lib/features/map_view/widgets/bottom_scroll_sheet/data_list.dart
```diff
@@ -16,19 +17,21 @@ class DataSliverList<T extends GoogleNavigable> extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     final itemsState = ref.watch(context.mapDataController<T>());
     return switch (itemsState) {
-      AsyncLoading() => const DataListLoading(),
+      // misz-masz here fixed flickering
+      // we probably should consider similar refactoring to all switches, but won't be visible in our other use-cases
+      AsyncValue(:final IList<T?> value) =>
+        _DataSliverList<T>(value.whereNonNull.toIList()),
```
mikolaj-jalocha, line 23: Does the introduction of immutable list reduced flickering?

lib/features/map_view/controllers/map_data_controller.dart
```diff
@@ -1,23 +1,22 @@
...
   String _textFieldFilterText = "";
   late final MapControllers<T> mapControllers;
 
   @override
-  FutureOr<Iterable<T?>?> build() async {
+  FutureOr<IList<T?>?> build() async {
     final itemSelected = ref.watch(mapControllers.activeMarker);
     if (itemSelected != null) {
-      return [itemSelected]; // shows only selected building
+      return [itemSelected].lock; // shows only selected building
```
mikolaj-jalocha, line 15: Tell me if i properly understand that; if the item is selected, the list cannot be modified?

mikolaj-jalocha (commented):
The flickering issue seemed simple to fix but ended up requiring quite a refactor. Good job! I wouldn’t have come up with a solution so quickly—my Flutter experience is too limited and my state management skills aren’t strong enough for that. 😂

Below are a few suggestions and some questions I have about how certain parts of the code work, so I can better understand the issue.

I really like the introduction of IList. It's an order of magnitude faster than Flutter's built-in version, and I think it could have a noticeable impact on our app’s performance. 
By the way, every time I review your code, I end up reading documentation on something new—which is cool!

lib/features/map_view/widgets/bottom_scroll_sheet/data_list.dart
simon-the-shark: It was one of many things I've tried cause `IList` implements list comparability, meaning `ILists` are equal if have equal elements in same order. Hovewer I forgot the riverpod doesn't check equality `==` operator anyway, but uses `identical` comparator. More you can read here in one of the questions https://riverpod.dev/docs/essentials/faq

It allowed me to reduce flickering frequency with manual check, but didn't eliminate it fully:
```dart
if (state != newstate) {
state = newstate; // or ref.invalidate or smth, basically reduces flickering to only situation when the number of places changed
} 
```



So its not the solution but I left the `IList` anyway cause it's recommended to have immutable and combarable state objects

lib/features/map_view/controllers/map_data_controller.dart
simon-the-shark: I believe `.lock` is just a shortcut for `.toIList()` but I didn't dive deeply

lib/features/map_view/widgets/map_widget.dart
simon-the-shark: Yeah we can use IList here

I moved the state listening down the tree in few places, meaning less widgets are refreshed needlesly. But it still doesn't help :P with the flickering

lib/features/map_view/widgets/bottom_scroll_sheet/data_list.dart
simon-the-shark: That's the change that helped. This alone solves the problem. I got the idea from docs as well, from the pull to refresh guide. https://riverpod.dev/docs/case_studies/pull_to_refresh

What was the problem is the fact that after refresh the state changes to `AsyncLoading` and we showed the loading screen for a second.

It turns out through, even though the state is loading. State.value still holds previous value for us to use.

So I changed the pattern matching switch logic a bit.

Now if the async value have any value we show it. If it has error we show error and in all other cases (`_` means all unmatched previously things) we show loading screen. So the loading screen shows if we have async loading or async data with value=null

simon-the-shark:
Thanks for the review! Hopefully i answered your questions here. If not the hit me up or we can hop on a call too.

Basically for state management i reccomend the official riverpod.dev docs. They are crazy good. I like to read all the pages there from time to time when I have a problem.

And also if youre not that familiar with pattern matching concept you can check it out too. Its very powerfull tool, pretty new to dart and we're using it in a very limited range here

PR #167, opened 2024-08-26
Fix/125 fixabout us center icons

mikolaj-jalocha:
I think section looks good ;) 

PS, imo it's a high time to gather pictures and social links from our Team to upload some content. I already typed names and roles, need pictures and links (it's good to take img size into consideration so that loading time is not exceptionally long)

Screenshot


simon-the-shark (approved):
About the image size. Directus allows us to request images in smaller size: https://learndirectus.com/image-optimization-in-directus/

So actually we can make a widget that will wrap our image with `LayoutBuilder` and always request proper size or something

Issue #124, opened 2024-08-03
fix/ parking typos


simon-the-shark:
I guess we should add "ul." to adresses? what you think?
<img width="345" alt="Zrzut ekranu 2024-08-3 o 15 51 38" src="https://github.com/user-attachments/assets/5bf8fe72-75c8-4871-a499-466cb3b76405">

mikolaj-jalocha:
agree

PR #168, opened 2024-08-26
#124 - parking typos - fix

mikolaj-jalocha:
At first glance, I thought about setting the street address in the data layer, but then I thought about localization and finally ended up setting the "ul. "  or "st. " prefix in the ui layer. 

Screenshot


simon-the-shark (approved):
we have a util to watch locale changes using `ref` and not `context`, but here I think what you've done is def enough

PR #177, opened 2024-08-29
Feat/ add wiredash feedback form

simon-the-shark:
I used wiredash for feedback form cause they have this really cool features as you can see in the video. 

they have 100k unique devices per month for free so no problem there.

they also have some really simple analytics. But we're gonna need to add Firebase Crashlytics and performance monitor anyway, so we can add analytics there too i guess. But for feedback form this looks imo pretty nice.

## analytics
Zrzut ekranu 2024-08-29 o 23 29 13

## feedback form

https://github.com/user-attachments/assets/6d4d2814-3126-4995-a902-717264d72918


lib/features/analytics/wiredash/show_feedback_tile.dart
```diff
@@ -0,0 +1,42 @@
...
+                id: "label-a9v91o7kdq",
+                title: context.localize.new_feature,
+              ),
+              Label(id: "label-bekp7mkn82", title: context.localize.bug),
+              Label(id: "label-ykl4o5vddu", title: context.localize.praise),
+            ],
+          ),
+        );
+      },
+      trailing: Lottie.asset(
```
mikolaj-jalocha, line 32: I'd consider using [renderCache ](https://pub.dev/packages/lottie#performance-or-excessive-cpugpu-usage)parameter here. For frame rate then you can use static variable that's stored in config

lib/features/analytics/wiredash/show_feedback_tile.dart
```diff
@@ -0,0 +1,42 @@
...
+  Widget build(BuildContext context) {
+    return WideTileCard(
+      title: context.localize.bug_report_title,
+      subtitle: context.localize.bug_report_subtitle,
+      onTap: () {
+        Wiredash.of(context).show(
+          options: WiredashFeedbackOptions(
+            labels: [
+              Label(
+                id: "label-a9v91o7kdq",
```
mikolaj-jalocha, line 23: I'm just wondering, if we can transfer all the labels ids to config file and name them appropriately?

lib/config/env.dart
```diff
@@ -14,11 +14,15 @@ abstract class ApiBaseConfig {
   useConstantCase: true,
   requireEnvFile: true,
 )
-abstract class ApiBaseEnv {
+abstract class Env {
```
mikolaj-jalocha, line 17: Like the change actually

mikolaj-jalocha (commented):
Report feedback mechanism looks great right now! Just like you said - very little coding here, which is good in my opinion.

I left few suggestions of mine ;)

lib/features/analytics/wiredash/show_feedback_tile.dart
simon-the-shark: cool

lib/features/analytics/wiredash/show_feedback_tile.dart
simon-the-shark: sure

PR #180, opened 2024-08-30
feat: add android adaptive icon

simon-the-shark:
as an android spec you can check out if anything seems suspicious in my icon setup :)
Zrzut ekranu 2024-08-30 o 22 29 49
but it does seems to work as expected


mikolaj-jalocha (approved):
LGTM! According to docs.

Issue #179, opened 2024-08-30
fix/ home screen offline refresh flickering


simon-the-shark:
when i click on refresh, it flickers. Probably the same cause as in parking flickering: #129 #152 

https://github.com/user-attachments/assets/25d5232c-bbab-4153-8b21-284d25555b03

simon-the-shark:
it's actually something with grapqh cache options

Issue #161, opened 2024-08-12
reafactor/ change all `riverpod`s provider that return `List` to return `IList`


simon-the-shark:
https://pub.dev/packages/fast_immutable_collections

Issue #142, opened 2024-08-05
A RenderFlex overflowed by 78 pixels on the bottom in Department's detail screen. 


mikolaj-jalocha:
After some previous fixes when entering department's details screen RenderFlex error is thrown. 

![image](https://github.com/user-attachments/assets/8a3c1441-591e-4020-b262-509d2a4ef1f9)

simon-the-shark:
So yeah, I guess we can return to previous layout or adjust department view to the new one as after mergin #181

PR #181, opened 2024-09-01
Fix/home screen offline flickering

### LINKED ISSUE #104: iOS scrolling down - strange behaviour

simon-the-shark:
Changes in chronological order:
1. changed fetch policy of graphql library from `cacheAndNetwork` to `onlyNetwork` in situation when we want to invalidate cache 
     - `cacheAndNetwork` caused that we saw stale data when we hit "refresh" when something was offline. (see video here #179 ) 
     - Now we see loading instead in this case
     - this kinda solved #179 - but stil the behavior is not ideal.
2. I added 300ms additional wait before throwing offline error
   - this way we can see loading for a while when we hit refresh and before this way less smoother - we'd seen it for a one frame, which also was pretty flickery I think
4. This allowed me to refactor all repositories from stream based to simpler/normal future based (cause now we only either use `onlyCache` or `onlyNetwork` polices.) 
     - stream based repos was kinda leftover of our setup before we had TTL of the cache.
     - this is simpler I think? Unless you prefer streams then we can get back, cause stream ones still work
5. Also added `KeepAliveHomeViewProviders` widget - which actually solves ANY problem of jumping scroll in any changes. 
     - So it also fixes #104 in different way
     - So if we want we can return to previous loading ui of `BigPreviewCard` - idk which you think is nicer, current or previous?

6. I also accumulated `Greeting` and `ExamSessionCountdown` into one widget cause they use the same repo, but previously were showing the same offline message twice.
7. This actually naturally led to `IList` refactoring #161 . But this is done in the next PR: #182  - this one is already pretty heavy I think


lib/api_base/query_adapter.dart
```diff
@@ -0,0 +1,46 @@
...
+
+  static Future<QueryResult<T>> graphqlTryCatch<T>(
+    QueryResult<T> event,
+    TtlKey ttlKey,
+  ) async {
+    if (!event.hasException) return event;
+
+    if (event.exception?.linkException != null) {
+      // wait, just to show loading on the ui for a little while
+      await Future.delayed(const Duration(milliseconds: 300));
```
mikolaj-jalocha, line 20: I like the change here

lib/features/academic_calendar/widgets/academic_calendar_consumer.dart
```diff
@@ -0,0 +1,45 @@
...
+import "package:flutter_riverpod/flutter_riverpod.dart";
+
+import "../../../config/ui_config.dart";
+import "../../../widgets/my_error_widget.dart";
+import "../../home_view/widgets/loading_widgets/horizontal_rectangular_section_loading.dart";
+import "../repository/academic_calendar_repo.dart";
+import "countdown_widget/exam_session_countdown.dart";
+import "home_screen_greeting.dart";
+
+class AcademicCalendarConsumer extends ConsumerWidget {
```
mikolaj-jalocha, line 11: I like naming convention here. Should all consumer widgets be named that way in other part of the project? Even private ones?

mikolaj-jalocha (approved):
LGTM! 

Some of my thoughts below: 

- _So if we want we can return to previous loading ui of BigPreviewCard - idk which you think is nicer, current or previous?_

    -  I think it doesn't really matter, both look nice imo. But I guess returning to previous one'd solve [this ](https://github.com/Solvro/mobile-topwr/issues/142)issue.

- stream based repos was kinda leftover of our setup before we had TTL of the cache.
this is simpler I think? Unless you prefer streams then we can get back, cause stream ones still work

     - I really like that change. It significantly reduce some of complexities with using Streams I guess. The project's api doesn't need to be constantly watched, so Future imo does the job perfectly :-)

lib/features/academic_calendar/widgets/academic_calendar_consumer.dart
simon-the-shark: I mean, I'm not enforcing it. I was just natural in this case. But yeah, if you want we can introduce some naming convention like this. But not to all `ConsumerWidget`s but to those that actually consumer some state - why not. Cause sometimes we also extend `ConsumerWidget` just to get `ref` for something

PR #187, opened 2024-09-04
Feat/ add days to semester mode

mikolaj-jalocha:
## About

- During the holiday season, the counter counts down the days until the start of the next semester.
- During the exam session, the counter tracks the days remaining until the session ends.
- After the semester starts and before the session begins, the counter displays the days between these two events.
- It's crucial to regularly update the dates in the API.

### DIFF
--- a/lib/features/academic_calendar/utils/counter_digits.dart
+++ b/lib/features/academic_calendar/utils/counter_digits.dart
@@ -7,13 +7,15 @@ import "../repository/academic_calendar_repo.dart";
 
 enum Digit { first, second, third }
 
-extension DaysLeftStringConverter on AcademicCalendar {
+extension DaysLeftStringConverterX on AcademicCalendar {
   String get daysLeftFromNowString {
     final calendarData = this.AcademicCalendarData;
     final daysLeft = calendarData != null
         ? (calendarData.isHolidays()
             ? data?.semesterStartDate.daysLeftFromNow
-            : data?.examSessionLastDay.daysLeftFromNow)
+            : calendarData.isExamSession()
+                ? data?.examSessionLastDay.daysLeftFromNow
+                : data?.examSessionStartDate.daysLeftFromNow)
         : null;
 
     return max(_default, daysLeft ?? _default)
@@ -22,7 +24,7 @@ extension DaysLeftStringConverter on AcademicCalendar {
   }
 }
 
-extension GetDigitExtension on String {
+extension GetDigitX on String {
   String getDigit(Digit digit) =>
       length > digit.index ? this[digit.index] : _default.toString();
 }
--- a/lib/features/academic_calendar/utils/counter_digits.dart
+++ b/lib/features/academic_calendar/utils/counter_digits.dart
@@ -2,18 +2,24 @@ import "dart:math";
 
 import "../../../config/ui_config.dart";
 import "../../../utils/datetime_utils.dart";
+import "../model/academic_calendar_extensions.dart";
 import "../repository/academic_calendar_repo.dart";
 
 enum Digit { first, second, third }
 
 extension DaysLeftStringConverter on AcademicCalendar {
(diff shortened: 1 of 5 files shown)


lib/features/academic_calendar/utils/localize_counter_message.dart
```diff
@@ -0,0 +1,17 @@
+import "package:flutter/cupertino.dart";
+
+import "../../../utils/context_extensions.dart";
+import "../model/academic_calendar_extensions.dart";
+import "../repository/academic_calendar_repo.dart";
+
+extension LocalizeCounterMessageExtension on AcademicCalendar {
+  String localizeDaysCounterMessage(BuildContext context) {
+    if (this.AcademicCalendarData!.isHolidays()) {
```
simon-the-shark, line 9: I don't like the non null assertion `!` here :) maybe better to be safe than sorry?

lib/features/academic_calendar/utils/counter_digits.dart
```diff
@@ -2,18 +2,24 @@ import "dart:math";
...
-        data?.examSessionStartDate.daysLeftFromNow ?? _default,
-      ).toString().padLeft(
-            Digit.values.length,
-            _default.toString(),
-          );
+  String get daysLeftFromNowString {
+    final calendarData = this.AcademicCalendarData;
+    final daysLeft = calendarData != null
+        ? (calendarData.isHolidays()
+            ? data?.semesterStartDate.daysLeftFromNow
```
simon-the-shark, line 15: So this basically here have only two modes/cases? Am I right? Shouldn't there be three cases as in you PR description? I think you don't have the case when we are in semester and we want to see days until the START of the exam session? Or am I missing smth?

lib/features/academic_calendar/utils/localize_counter_message.dart
```diff
@@ -0,0 +1,17 @@
+import "package:flutter/cupertino.dart";
+
+import "../../../utils/context_extensions.dart";
+import "../model/academic_calendar_extensions.dart";
+import "../repository/academic_calendar_repo.dart";
+
+extension LocalizeCounterMessageExtension on AcademicCalendar {
```
simon-the-shark, line 7: missing X in the name I guess?

lib/features/academic_calendar/utils/localize_counter_message.dart
```diff
@@ -0,0 +1,17 @@
...
+  String localizeDaysCounterMessage(BuildContext context) {
+    if (this.AcademicCalendarData!.isHolidays()) {
+      return context.localize.until_the_end_of_the_semester_break;
+    }
+    if (this.AcademicCalendarData!.isExamSession()) {
+      return context.localize.until_the_session_ends;
+    }
+    return context.localize.days;
```
simon-the-shark, line 15: should here be `to_start_session` label instead?

simon-the-shark (commented):
I like the whole idea. Not sure if you didn't miss the initial case of when you count days until the start of exam session. But maybe I've missing something

lib/features/academic_calendar/utils/localize_counter_message.dart
mikolaj-jalocha: Sure, I also refactored other parts of code here that didn't use that naming convention

lib/features/academic_calendar/utils/counter_digits.dart
mikolaj-jalocha: Sorry for that. Of course you're right here.  I think's fixed rn.

mikolaj-jalocha:
> I like the whole idea. Not sure if you didn't miss the initial case of when you count days until the start of exam session. But maybe I'm missing something

Sorry for that, a bit of amateur here.. 

I guess everything's fixed right now. Please take o look ;)

simon-the-shark (approved):

PR #195, opened 2024-09-05
Feat/place fav parking on top of parkings list

TTomaszPWR:


lib/features/parkings_view/repository/parkings_repository.dart
```diff
@@ -26,6 +27,26 @@ Future<IList<Parking>> parkingsRepository(ParkingsRepositoryRef ref) async {
...
+List<Parking> _sortParkingsByFav(
+  List<Parking> list,
+  ParkingsRepositoryRef ref,
+) {
+  final List<Parking> finalParkings = [];
+  for (final parking in list) {
+    final isFavorite =
+        ref.watch(localFavParkingsRepositoryProvider(parking.id)) ?? false;
+    if (isFavorite) {
+      finalParkings.insert(0, parking);
```
simon-the-shark, line 46: I know this talk has 0 sense when we have only 5 parkings and pwr probably won't build next 1000 or 10k in a night - but when I see inserting elements at 0 index of an ArrayList in a loop, I do see this as a redflag :)

On a big datasets this will be pretty heavy operation - maybe you remember this from java classes from uni (xd). I assume you more or less know what I mean, but if you want deeper explanation, I'm more than happy to explain this.

Default Dart's list is some sort of array list implementation.

And you only insert at the beggining and the end and then right away use `toIList` conversion, which iterates over collection 1 time to create a completely different data structure - so this list doesn't need random index access.

IMO this seems as a perfect occasion to use some sort of Linked List implementation, e.g. `DoubleLinkedQueue` from `collections` package. We already use this package and this should be a pretty quick change. 

But for this size of dataset your solution is more than valid, so you can leave it too if you want it.

simon-the-shark (commented):
Seems to work for me. Left one quick note - you can disagree. 

I also do see here a occasion for next small finishing task for this feature - maybe you'll be interested: I would investigate options to animate this list to make changing the fav state smoother. (there seems to be some ready to use packages but I've never used one like this)

TTomaszPWR:
> Seems to work for me. Left one quick note - you can disagree. 
> 
> I also do see here a occasion for next small finishing task for this feature - maybe you'll be interested: I would investigate options to animate this list to make changing the fav state smoother. (there seems to be some ready to use packages but I've never used one like this) (https://github.com/Solvro/mobile-topwr/issues/197)

I think it is a great idea. I can do it. Should I merge this or continue changes on this branch?

simon-the-shark:
Just change the list and we'll merge and animation will be a next task: https://github.com/Solvro/mobile-topwr/issues/197

simon-the-shark (approved):

PR #207, opened 2024-09-15
Feat/ add remote cache flushing

simon-the-shark:
we check remote number on directus and if it increased, we flush all cache

tell me what you think

### DIFF
--- a/lib/api_base/cache/remote_flush/business/flush_cache_use_cases.dart
+++ b/lib/api_base/cache/remote_flush/business/flush_cache_use_cases.dart
@@ -1,0 +1,29 @@
+import "dart:async";
+
+import "package:flutter_riverpod/flutter_riverpod.dart";
+
+import "../../../../config/ttl_config.dart";
+import "../../ttl/ttl_service.dart";
+import "../data/local_cache_ref_number_repo.dart";
+import "../data/remote_cache_ref_number_repo.dart";
+
+extension FlushCacheUseCasesX on WidgetRef {
+  Future<void> flushAllCacheIfNeeded() async {
+    final remoteRefNumber =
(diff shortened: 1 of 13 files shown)


mikolaj-jalocha (approved):
LGTM! 

I've one question. In Android native development I'd probably use one of the lifecycle 's method. What about [them ](https://yawarosman.medium.com/flutter-lifecycle-methods-an-in-depth-exploration-00b0747864bb)here?

simon-the-shark:
@mikolaj-jalocha  Do you want to use **app's** lifecycle method? Cause I already used `initState` which is **widget's** lifecycle method and it will be triggered only on first appereance of the widget in the tree, which in this cause (taking into account its position in the tree) - means it makes the request only once, after opening the app. You need to close the app and reenter to retrigger the request. So do you think we should  check this also when we get back to the app from the background or what's your specific proposition?

mikolaj-jalocha:
> @mikolaj-jalocha Do you want to use **app's** lifecycle method? Cause I already used `initState` which is **widget's** lifecycle method and it will be triggered only on first appereance of the widget in the tree, which in this case (taking into account its position in the tree) - means it makes the request only once, after opening the app. You need to close the app and reenter to retrigger the request. So do you think we should check this also when we get back to the app from the background or what's your specific proposition?

No no, everything here works as expected. I was wondering if  we could  use only lifecycle methods and ommit creating separate widget. Just curious 🤔

simon-the-shark:
We can use `WidgetsBindingObserver`, but it's a mixin for a widget (so we need widget) or we can use `AppLifecycleListener` class, which is a separate class (nice to use) - but we still need to attach it to the app (and we need to do that in some widget :P) - so we do need widget. If we use hooks, this can be simplified (less boilerplate)

PR #208, opened 2024-09-15
refactor/replace dashes with arrows trend and fix layout

TTomaszPWR:


TTomaszPWR:
I changed padding in some places because chart didn't work proper

simon-the-shark (commented):
nicely done, no bigger notes. It looks nice too, but I would use more consistent arrows (now down & right arrows look different)
image


# my proposition
I did a quick research and only consistent set of arrows in default set are the compass directions icons:
```dart
Icons.east,
Icons.north_east,
Icons.south_east,
```
Zrzut ekranu 2024-09-15 o 18 10 40

so my proposition is to use this arrow set instead - what you think? apart from this, it looks good to me 😎

TTomaszPWR:
> nicely done, no bigger notes. It looks nice too, but I would use more consistent arrows (now down & right arrows look different) image
> 
> # my proposition
> I did a quick research and only consistent set of arrows in default set are the compass directions icons:
> 
> ```dart
> Icons.east,
> Icons.north_east,
> Icons.south_east,
> ```
> 
> Zrzut ekranu 2024-09-15 o 18 10 40
> so my proposition is to use this arrow set instead - what you think? apart from this, it looks good to me 😎

I think it is a good idea

simon-the-shark (approved):

Issue #199, opened 2024-09-05
try to use svg optimizer


simon-the-shark:
not sure if it's worth it and a good idea, but maybe we can try this out
- https://www.thedroidsonroids.com/blog/svg-open-source-flutter-package

PR #211, opened 2024-09-17
Try to use svg optimizer

mikolaj-jalocha:
- as described [here](https://github.com/Solvro/mobile-topwr/issues/199)
- since our project has under 20 svg files difference in performance's probably minor (below 1%) but imo still worth implementing
- since svg_optimizer uses SVGO under the hood, Node.js must be installed locally


pubspec.yaml
```diff
@@ -71,6 +71,7 @@ dependencies:
   package_info_plus: ^8.0.2
   animated_list_plus: ^0.5.2
   flutter_hooks: ^0.20.5
+  svg_optimizer: ^0.0.1
```
simon-the-shark, line 74: did you try to use it with it in `dev dependencies`

pubspec.yaml
```diff
@@ -106,6 +107,9 @@ flutter:
   generate: true
 
   assets:
+    - path: assets/
```
simon-the-shark, line 110: this actually increases app bundle size, cause it includes eveyrything in the assets folder, and we do not use everything there (e.g. banner). Maybe let's try to place all svgs in /assets/svgs ale set this path here. Let's see how this affects the size?

simon-the-shark (commented):
yeah, now i think this could be an overkill. But maybe we can try this with the note I've mentioned

simon-the-shark (approved):
Yeah, I don't see much change in the bundle size, but let's go with it anyway

Issue #198, opened 2024-09-05
feat/ add proper deep links / app links


simon-the-shark:
https://codewithandrea.com/articles/flutter-deep-links/

simon-the-shark:
The above tutorial is more about go router, but it should be similar to auto route. auto route also have some info in docs: https://pub.dev/packages/auto_route#deep-linking

simon-the-shark:
Maybe not redirecting from our website, but more like we would want to have a possibility to send a link that will open the app at e. g. specific departament details or specific guide section

mikolaj-jalocha:
@simon-the-shark 

Commit looks extremely simple but trust me, it took me 3h to set it up accordingly XD.  Since I'm not that rich yet, I haven't got a Mac machine. Could you please set up iOS config regarding deeplinks? I belive proper docs could be found [here](https://docs.flutter.dev/cookbook/navigation/set-up-universal-links#adjust-ios-build-settings).

**Things to know before:**

- default domain is: topwr.solvro.pl  eg.: topwr.solvro.pl/guide   topwr.solvro.pl/departments/1  etc.
- uri's end must be same as path set up in navigation. In our project paths are meaningful and self-descriptive so I didn't set additional config as described [here](https://pub.dev/packages/auto_route#deep-linking). If the uri's end is other we must perform additional coding but I believe default paths are enough (at least for now). 


- Android solution's been tested on my real device (Xiaomi Mi9 Se) and works fine.  Please, test it on your own as well.  
- What's worth mentioning, when link's opened within eg. Messenger, app web view opens (not our app), same happens with deep links from other apps like OLX. It seems that we can't do much about it.

simon-the-shark:
Sure, I'll look into it

simon-the-shark:
@mikolaj-jalocha you can make draft PR though

mikolaj-jalocha:
> @mikolaj-jalocha you can make draft PR though

sure, done

PR #215, opened 2024-09-18
#198 - deep links support - android setup

mikolaj-jalocha:
@simon-the-shark

Commit looks extremely simple but trust me, it took me 3h to set it up accordingly XD. Since I'm not that rich yet, I haven't got a Mac machine. Could you please set up iOS config regarding deeplinks? I belive proper docs could be found [here](https://docs.flutter.dev/cookbook/navigation/set-up-universal-links#adjust-ios-build-settings).

Things to know before:

default domain is: topwr.solvro.pl eg.: topwr.solvro.pl/guide topwr.solvro.pl/departments/1 etc.

uri's end must be same as path set up in navigation. In our project paths are meaningful and self-descriptive so I didn't set additional config as described [here](https://pub.dev/packages/auto_route#deep-linking). If the uri's end is other we must perform additional coding but I believe default paths are enough (at least for now).

Android solution's been tested on my real device (Xiaomi Mi9 Se) and works fine. Please, test it on your own as well.

What's worth mentioning, when link's opened within eg. Messenger, app web view opens (not our app), same happens with deep links from other apps like OLX. It seems that we can't do much about it.

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -13,7 +13,7 @@
             android:name=".MainActivity"
             android:enableOnBackInvokedCallback="true"
             android:exported="true"
-            android:launchMode="singleTop"
+            android:launchMode="singleTask"
             android:theme="@style/LaunchTheme"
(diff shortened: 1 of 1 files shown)


mikolaj-jalocha:
What about link that ends: \departments\1?

Issue #205, opened 2024-09-14
feat/ add shimmer loading to about us


simon-the-shark:
<img width="289" alt="Zrzut ekranu 2024-09-14 o 09 46 59" src="https://github.com/user-attachments/assets/d5b649a0-5561-4347-8592-d7b3acb946e5">

simon-the-shark:
we can prolly reuse the loading effect from guide detail view

PR #224, opened 2024-09-22
feat: add shimmer effect on about us view

kamilfedio:
#205 
I think is better way to write some shimmer boxes than using the same ui style us in guide view for user experience. If u think differently I can change it ofc

### DIFF
--- a/lib/features/about_us_view/about_us_view.dart
+++ b/lib/features/about_us_view/about_us_view.dart
@@ -6,7 +6,9 @@ import "../../config/ui_config.dart";
 import "../../utils/context_extensions.dart";
 import "../../widgets/detail_views/detail_view_app_bar.dart";
 import "../../widgets/detail_views/sliver_header_section.dart";
+import "../../widgets/loading_widgets/scrolable_loader_builder.dart";
 import "../../widgets/loading_widgets/shimmer_loading.dart";
+import "../../widgets/loading_widgets/simple_previews/preview_card_loading.dart";
 import "../../widgets/my_error_widget.dart";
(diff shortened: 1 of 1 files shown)


mikolaj-jalocha:
> #205 I think is better way to write some shimmer boxes than using the same ui style us in guide view for user experience. If u think differently I can change it ofc

Could you please attach some screenshots?

kamilfedio:
@mikolaj-jalocha yup
image

simon-the-shark (approved):
Good job!

My previous suggestion about using guide view was just a thing to make this task quicker - but customized view is even better. So good job here. It looks fire 🔥 

The code is fine as well - my only note is that you could have made your work a bit easier using our ready to use helpers like `PreviewCardLoading` or `ScrollableLoaderBuilder` in few places. But it's not a problem for me, it's just a loading view so your build is actually maybe even a bit more performant like this and it looks fine too. So if you want you can check if you can use `PreviewCardLoading` or `ScrollableLoaderBuilder` or smth here, but we can also just merge it honestly. So you can try to refactor if you want or we're merging it - whatever you prefer - just let me know. I do give merge approval here even with the current state of things

kamilfedio:
Okay I will try to refactor tommorow

kamilfedio:
@simon-the-shark  okay little delay but i refactor. Only im not sure aboutmy correct use of `ScrollableLoaderBuilder`

simon-the-shark (approved):
Great, You've used it correctly. My only note is that these widgets are self sufficient and you do not any extra Shimmer or ShimmerLoadingItem above it

PR #225, opened 2024-09-22
refactor/ try flutter hooks

simon-the-shark:
So this refactor is **optional** - if you don't like it, we don't need to merge it. Everything should work exactly the same without it. 

So hooks are basically thing borrowed from React and I was previously skeptical as they don't feel 100% Fluttery in a way. But once you get used to them, they have much less boilerplate, they are very consice (and statefull widget is not really). Also you can very easily write reusable logic (by creating your own hooks - by just writing a function). You can read more about hooks e.g. here https://riverpod.dev/docs/concepts/about_hooks or here https://pub.dev/packages/flutter_hooks

Also they are reccomended by riverpod as solution for "local state" and riverpod should be more about global/bussiness state.

But again, this PR is purely refactor and should change any behaviour. We can skip it or merge it, tell me what you think


lib/hooks/use_effect_on_init.dart
```diff
@@ -0,0 +1,8 @@
+import "package:flutter_hooks/flutter_hooks.dart";
```
mikolaj-jalocha, line 1: So, instead of that function, we alternatively could use default `useEffect` function always passing empty list? And since we often need to run some state related function once, on init, this wrapper is handy. Am I right? 
If yes, it is nice improvement actually. Hooks got my heart.

lib/widgets/search_box.dart
```diff
@@ -1,67 +1,69 @@
...
+    final controller = useTextEditingController();
+    final showCloseIcon = useState(false);
 
-  void _onChanged(String query) {
-    setState(() {
-      showCloseIcon = query != "";
-    });
-    widget.onQueryChanged(query);
-  }
+    final onTapOutside = useCallback(
```
mikolaj-jalocha, line 27: Follow up to previous question (about `useEffectOnInit`), the callback here'd be executed the first time the hook is invoked, and then whenever focus node changes (meaning the whole list changes)?

mikolaj-jalocha (approved):
I got in love with hooks from the first line of this review ❤️

Code now got a lot more readable and easier to understand (especially for someone not deep or experienced in flutter). Tbh code now resembles Android native in some ways of managing state. 

I like the idea and opt for it with all my heart in it. I leave some questions of mine

lib/hooks/use_effect_on_init.dart
simon-the-shark: exactly - we can use `useEffect` with empty dependency list everytime, but we can easily create custom hooks like this in very simple way to introduce reusability. This is very simple example of a custom hook, but custom hooks can also have a big chunks of logic too

lib/widgets/search_box.dart
simon-the-shark: so there is hook called `useMemoized` which caches some expensive operation/creation of sometiming and recalculates only when someting in the dependecy list changes

simon-the-shark: and `useCallback` is a simple wrapper on `useMemoized` but for functions

simon-the-shark: this basically means it creates this function inside only first time and then on rebuild, it doesn't create this function again but uses previous instance - unless a key in dependecy list has changed

mikolaj-jalocha: > this basically means it creates this function inside only first time and then on rebuild, it doesn't create this function

I thin I got it now! Thank you

PR #226, opened 2024-09-22
Fix/ scope providers to specific routes

simon-the-shark:
so if you check out the previous PR, about hooks, I've mentioned there, that riverpod is more recomened for global state only and not "emphermal"/local state. And we were using it 100% right at the start, but now with new navigator setup, this is not 100% true anymore. Cause if we opened same tab multiple times, some of the state was shared due to global riverpod's providers. 

So riverpod docs really recommend hooks for this purpose. In our case it's text fields' contents and selected filters - these are more like local state and not business/global one.

But I did try to use hooks for this and didn't like the results fully there. I went for a quicker fix of this, which is riverpod scoping: https://riverpod.dev/docs/concepts/scopes. Scoping on the other hand is not my favourite mechanism, cause it's hard to debug - but once you manage to set up this correctly, this works pretty ok.

So this is a quick refactor/fix that scopes riverpod to their routes only. This means we cannot change their state outside the mentioned route it's scoped to. I mean, we can  - but we won't see the effect we wanted. That's why this can be messy to debug. 

In the future, we can try to make some refactor and use e.g. hooks for all the text field's states and selected filters states. But this is a bigger refactor to make it work nice. And this one is a quick fix that works nicely. The only change is that we need to remember the state are scoped - so I made another PR that will fix few this PR has broke few things now (nothing serious)


mikolaj-jalocha (approved):
I think I generally understand concept for that. 

Previously, due to our nav bar setup, some of the (local) state was shared globally, I can see the danger of it (and I agree totally that's not desired behawior and it's excellent that this PR fix this) but I wonder if it caused us some real issues before, or it could one day?

simon-the-shark:
real issue could be I think related to text fields, cause text field itself also have its local state that is only synchronised with riverpod via `onChanged` - apart from this it was only issue that e.g. filters did not reset if you leave and reenter tab but not real isssues. Basically there were small issues if you played around really hard with the app :)

PR #227, opened 2024-09-22
Feat/active markers and filters deeplink support

simon-the-shark:
so this finally is a final PR, stacked on top of previous PRs - which finally adds some feature.

this add support for more deeplinks and also fixes previous cross-using scoped providers (thigs that previous PR broke)

So now we can have this deeplinks working too:
```
adb shell am start -a android.intent.action.VIEW \
  -c android.intent.category.BROWSABLE \
  -d https://topwr.solvro.pl/buildings/2  \
 com.solvro.topwr
```
```
adb shell am start -a android.intent.action.VIEW \
  -c android.intent.category.BROWSABLE \
  -d https://topwr.solvro.pl/parkings/2  \
 com.solvro.topwr
```

and even we can choose filters using query params:
```
adb shell am start -a android.intent.action.VIEW \
  -c android.intent.category.BROWSABLE \
  -d "https://topwr.solvro.pl/sci-clubs/\?tags=Turystyka,Sport\&depts=10,3,5,6\&types=scientific_cirlce,student_organization" \
  com.solvro.topwr
```

this seems to work but needs some testing (especially on IOS)

I don't think the code here is not the cleariest, but that's partially due to using riverpod for semi-local state too. So this could maybe be simplified a bit, if we used hooks for everything - but it works for now

tell me what you think


mikolaj-jalocha (approved):
Excellent, more precise deeplink support was needed. Nice feature actually. 

I totally agree that code in some places is hard to understand, as you said, due to riverpods specs...  As with previous PR, my experience in Flutter framework is to short for now to spot some other possible solutions  (as in some other lighter PR😂 ) or violations  but I didn't found anything catching my attention here ;) So good work, I guess 

I'd definitely opt here for hooks. Bit complex refactor, but I just thought that if you like and have enough time and willingness we could do it together on a call, maybe Tomek Trela or Kamil'd  be interested as well? This part of our app seems crucial and imo it's not the easiest to understand (due to navigation nature..) so maybe some codeshare'd simplify things a bit, and maybe introduce sth new at Solvro 🤔

simon-the-shark:
yeah we can do this refactor but not for this release. But yeah, I like the idea with call, but we need to find some free time

Issue #221, opened 2024-09-21
Map tile's caching with hive


simon-the-shark:
@simon-the-shark  just refactored map mechanism in our project. [Please see this PR](https://github.com/Solvro/mobile-topwr/pull/220). 

There's urgent need for caching map tiles. 

https://pub.dev/packages/flutter_map_cache

PR #229, opened 2024-09-25
#221 - add cache support for map tiles

mikolaj-jalocha:
I've added cachce support for map tiles. 

- I've been wondering if flushing map's cache should be possible from an API. I've decided not to implement that feature for now (maps don't change that often I guess,at least ones that we use).  If you think differently, I can change it of course
- in the Hive implementation you can see `null` passed in place of directory path. Documentation says that if we use Hive already in the project, there's no need to provide it again. I guess that's correct in our situation. 
- do you know any ways to change if caching actually works (some logging or sth)?

### DIFF
--- a/lib/features/map_view/widgets/map_widget.dart
+++ b/lib/features/map_view/widgets/map_widget.dart
@@ -1,6 +1,7 @@
+import "package:dio_cache_interceptor_hive_store/dio_cache_interceptor_hive_store.dart";
 import "package:flutter/material.dart";
 import "package:flutter_map/flutter_map.dart";
-import "package:flutter_map_cancellable_tile_provider/flutter_map_cancellable_tile_provider.dart";
+import "package:flutter_map_cache/flutter_map_cache.dart";
 import "package:flutter_riverpod/flutter_riverpod.dart";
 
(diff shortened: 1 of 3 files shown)


lib/features/map_view/widgets/map_widget.dart
```diff
@@ -38,7 +39,13 @@ class MapWidget<T extends GoogleNavigable> extends ConsumerWidget {
         TileLayer(
           urlTemplate: OpenStreetMapConfig.tileUrl,
           userAgentPackageName: OpenStreetMapConfig.userAgent,
-          tileProvider: CancellableNetworkTileProvider(),
+          tileProvider: CachedTileProvider(
```
simon-the-shark, line 42: future note, on web we may want to use CancellableNetworkTileProvider instead of cached one - but only a future note

simon-the-shark (approved):
LGTM!

1. TTL of map cache is def enough
2. yeah we use it indirectly with graphql package
3. so just install app, enter it and then exit app, enter flight mode (no internet) and if map has still tiles on new offline enter then the caching works. -imo simpelst testing method

Issue #230, opened 2024-09-25
Replace decimal numbers with integers at parking's page


mikolaj-jalocha:
![Image](https://github.com/user-attachments/assets/08172782-9811-4465-a4f0-33f9c22c5720)

PR #231, opened 2024-09-25
#230 - replace decimal numbers with integers at parking's page & make sure that free slots are always >= 0

mikolaj-jalocha:
- It would be perfect to notify the person reporting the error for morale support.  
  Image 1

- I’ve also implemented a mechanism that ensures that the number of available spaces is always equal to or greater than 0 (never negative, as happened on 24/09/2024).  
  Image 2


simon-the-shark (approved):
LGTM

PR #232, opened 2024-09-25
fix/overflow-in-sci-clubs-card

tomasz-trela:


simon-the-shark (changes requested):
sorry to say this but I don't think this is the way we want to handle this

Zrzut ekranu 2024-09-26 o 00 33 38

5 lines look much better but it has problems in few edge cases I agree. We can discuss how we want to handle that

TTomaszPWR:
> sorry to say this but I don't think this is the way we want to handle this
> 
> Zrzut ekranu 2024-09-26 o 00 33 38
> 
> 5 lines look much better but it has problems in few edge cases I agree. We can discuss how we want to handle that

Hmm, we can limit lines for science club name. I think it is a good idea. What do you think @simon-the-shark ?

mikolaj-jalocha:
> > sorry to say this but I don't think this is the way we want to handle this
> > Zrzut ekranu 2024-09-26 o 00 33 38
> > 5 lines look much better but it has problems in few edge cases I agree. We can discuss how we want to handle that
> 
> Hmm, we can limit lines for science club name. I think it is a good idea. What do you think @simon-the-shark ?

I'm not sure what's the exact problem but it's long names of organizations I suppose. I'd think about some kind of scalling long names. Imo they should be visible. Similar thing was done at home's screen section regarding sci clubs. You can check the solution there, maybe it'd suit for this as well.

TTomaszPWR:
> > > sorry to say this but I don't think this is the way we want to handle this
> > > Zrzut ekranu 2024-09-26 o 00 33 38
> > > 5 lines look much better but it has problems in few edge cases I agree. We can discuss how we want to handle that
> > 
> > 
> > Hmm, we can limit lines for science club name. I think it is a good idea. What do you think @simon-the-shark ?
> 
> I'm not sure what's the exact problem but it's long names of organizations I suppose. I'd think about some kind of scalling long names. Imo they should be visible. Similar thing was done at home's screen section regarding sci clubs. You can check the solution there, maybe it'd suit for this as well.

That's sounds better. I'll do it.

TTomaszPWR:
After many attempts to auto-size text. I've decided to use pub dev library.

lib/features/science_clubs_view/widgets/ensure_visible_tags.dart
```diff
@@ -46,10 +47,10 @@ class EnsureVisibleTags extends DualTextMaxLines {
         return Column(
           crossAxisAlignment: CrossAxisAlignment.stretch,
           children: [
-            RichText(
+            AutoSizeText.rich(
```
simon-the-shark, line 50: its' actually a real funny combination that we do calculate the max lines with our logic and then apply scaling down but the results looks pretty ok

simon-the-shark (approved):

simon-the-shark:
I'll maybe try to limit department to 1 line too - maybe the tags will be better looking. But merging this for now

Issue #240, opened 2024-09-27
fix/ unify bottom paddings among tab views


simon-the-shark:
![image](https://github.com/user-attachments/assets/cd7bc03f-e822-4f4b-ac1c-57e7ddc3ed5b)
![image](https://github.com/user-attachments/assets/ac2b579a-0caa-4c1d-94ac-9257a57d37ea)

![image](https://github.com/user-attachments/assets/34eb3a5d-3c17-454f-a31c-d9a86334159b)

PR #246, opened 2024-09-27
fix: change bottom padding on home view

kamilfedio:
#240 
I changed bottom padding on home view. Now all tabs have the same -> 24px
some ss's:
image
image
image

### DIFF
--- a/lib/features/home_view/home_view.dart
+++ b/lib/features/home_view/home_view.dart
@@ -32,7 +32,7 @@ class HomeView extends StatelessWidget {
       appBar: LogoAppBar(context),
       body: KeepAliveHomeViewProviders(
         child: ListView.separated(
-          padding: const EdgeInsets.only(bottom: 48),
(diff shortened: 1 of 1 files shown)


lib/features/home_view/home_view.dart
```diff
@@ -32,7 +32,7 @@ class HomeView extends StatelessWidget {
       appBar: LogoAppBar(context),
       body: KeepAliveHomeViewProviders(
         child: ListView.separated(
-          padding: const EdgeInsets.only(bottom: 48),
+          padding: const EdgeInsets.only(bottom: 24),
```
mikolaj-jalocha, line 35: Please declare it at `HomeViewConfig `:-)

mikolaj-jalocha (changes requested):

lib/features/home_view/home_view.dart
kamilfedio: done

Issue #238, opened 2024-09-27
feat/ sort links on about us team 


simon-the-shark:

wkrzos:
@Rei-x has told me he can take a look at it as well.

mikolaj-jalocha:
> @Rei-x has told me he can take a look at it as well and make changes to the directus setup.

Currently, sorting logic (eg. at `StudyCircleContactSection`) is implemented locally, it doesn't depend on api setup. Similar logic needs to be implemented in `About Us` section.

PR #249, opened 2024-09-27
feat - #238 sort links on about us team

mikolaj-jalocha:
#238 
- separate sorting is now included for sci clubs and about us team
- maybe it's good idea to make sorting not local but on backend? @Rei-x


Rei-x:
now it is sorted on the backend

simon-the-shark:
@mikolaj-jalocha i guess we can skip this PR now

PR #259, opened 2024-09-28
feature: add tooltip widget with config

kamilfedio:
#253 
I have added tooltip widget with some decoration
some ss's:
image


kamilfedio:
version with some box shadow since tooltip is unreadable
image
Let me know if it is enough <3

simon-the-shark (commented):
LGTM 😎

Issue #263, opened 2024-10-01
feat/add-verified-checkmark-to-science-clubs-on-homepage


wkrzos:
![Image](https://github.com/user-attachments/assets/29c46060-85c6-4026-a4cf-8a8e95c20e9b)

PR #291, opened 2024-10-04
Feat/ add verified checkmark to science clubs in all other places

mikolaj-jalocha:
#263 

- Checkmarks are now displayed in every place that study circle tile's present.
- I've refactored logic responsible for determining if checkmarks should be displayed.
- Please check the visual effect and tell if you like it ;)
- I wonder if we should implement some kind of checkmark even in the study circle's details?

### DIFF
--- a/lib/config/ui_config.dart
+++ b/lib/config/ui_config.dart
@@ -114,6 +114,7 @@ abstract class ScienceClubsViewConfig {
 
 abstract class ScienceClubCardConfig {
   static const trailingPadding = 2.0;
+  static const source = "manual_entry";
 }
 
 abstract class ParkingsConfig {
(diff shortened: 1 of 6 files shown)


lib/config/ui_config.dart
```diff
@@ -114,6 +114,7 @@ abstract class ScienceClubsViewConfig {
 
 abstract class ScienceClubCardConfig {
   static const trailingPadding = 2.0;
+  static const source = "manual_entry";
```
simon-the-shark, line 117: nice

simon-the-shark (approved):
Good job! Looks nice. I think we can add it the same way for the detail view next to the name. That's a good idea. Task tracker: [#292](https://github.com/Solvro/mobile-topwr/issues/292)

PR #294, opened 2024-10-05
Feat/ add app's streak counter

simon-the-shark:
Screenshot_1728146623

### DIFF
--- a/lib/features/app_streak/data/streak_start_repo.dart
+++ b/lib/features/app_streak/data/streak_start_repo.dart
@@ -1,0 +1,21 @@
+import "dart:async";
+
+import "package:riverpod_annotation/riverpod_annotation.dart";
+
+import "../../../config/shared_prefs.dart";
+import "../../../utils/timestamp.dart";
+import "timstamp_repo_mixin.dart";
+
+part "streak_start_repo.g.dart";
+
+@riverpod
+class StreakStartRepository extends _$StreakStartRepository
+    with TimestampRepository {
+  @override
+  final storeKey = StreakRepositoriesConfig.streakStartKey;
+
+  @override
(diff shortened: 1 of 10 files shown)


lib/features/app_streak/data/streak_start_repo.dart
```diff
@@ -0,0 +1,20 @@
...
+import "../../../utils/timestamp.dart";
+import "timstamp_repo_mixin.dart";
+
+part "streak_start_repo.g.dart";
+
+@riverpod
+class StreakStartRepository extends _$StreakStartRepository
+    with TimestampRepository {
+  @override
+  final storeKey = "__app_streak_counter__streak_start_stamp";
```
mikolaj-jalocha, line 14: What about parametrizing it in the config file?

lib/features/app_streak/presentation/app_streak_counter.dart
mikolaj-jalocha, line 1: Maybe an overkill, but I thought about adding simple loading shimmer here (it's everywhere else)

mikolaj-jalocha (approved):
LGTM! 

I quickly grasped what the code was about, so nice work. Everything nice & clean along with reusable mixin. I've left 2 comments.

lib/features/app_streak/presentation/app_streak_counter.dart
simon-the-shark: it should basically be almost instantenious - no actualy loading should be seen imo in this case (shared preferences should be already loaded)

lib/features/app_streak/data/streak_start_repo.dart
simon-the-shark: sure

PR #297, opened 2024-10-08
feat/add-shadows-to-trend-arrow

tomasz-trela:
Now trend arrow has shadows like every component in ParkingWideTileCard


mikolaj-jalocha:
LGTM!

Issue #286, opened 2024-10-04
Feat/ add in-app-review


simon-the-shark:
As a user I shall be asked to leave a review in store using in-app-review technology. We need to track some events and asks only if user is engaged (as a start we may use consecutive days counter #285)

PR #345, opened 2024-10-26
#286 feat add in app review

mikolaj-jalocha:
#286 

### User is prompted to rate our app only if:
- his day streak's at least 3
- total review prompt doesn't exceed 2
- last prompt was displayed at least 7 days ago

Testing on Android is quite difficult (please refer to [this](https://pub.dev/packages/in_app_review#testing-read-carefully)). Google Play access's needed to test if behawior works as expected so I couldn't empirically check if everything's fine on Android :-)    Since I'm not that rich yet hence don't own a Mac, iOS testing couldn't be done by me as well :-) . So if you can, please test it. You can contact me when testing on Android (I'm curious about how the internal app sharing process looks like)

### DIFF
--- a/lib/features/in_app_review/business/in_app_rating_service.dart
+++ b/lib/features/in_app_review/business/in_app_rating_service.dart
@@ -1,0 +1,59 @@
+import "package:flutter_riverpod/flutter_riverpod.dart";
+import "package:in_app_review/in_app_review.dart";
+import "package:shared_preferences/shared_preferences.dart";
+
+import "../../../config/shared_prefs.dart";
+import "../../../utils/datetime_utils.dart";
+import "../../app_streak/business/get_days_use_case.dart";
+
+
+class InAppRatingService {
+  const InAppRatingService(this.ref);
+
+  final Ref ref;
+
+  InAppReview get _inAppReview => InAppReview.instance;
+
+  Future<int> _getUsageDays() async {
+    return await ref.read(getUsageDaysUseCaseProvider.future);
+  }
+
+  SharedPreferences get _sharedPrefs =>
+      ref
+          .read(sharedPreferencesSingletonProvider)
+          .requireValue;
+
+  Future<void> requestReviewIfNeeded() async {
+    if (await _canRequestReview()) {
+      await _inAppReview.requestReview();
+      await _updatePreferences();
+    }
+  }
+
+  Future<bool> _canRequestReview() async {
+    final lastReviewPrompt = _sharedPrefs.getInt(
+        InAppReviewConfig.lastReviewPromptDateInMillis,);
+    final reviewCount = _sharedPrefs.getInt(InAppReviewConfig.reviewCountKey) ??
+        0;
+    final usageDays = await _getUsageDays();
+
+    if (usageDays > 2 && reviewCount < 3 && await _inAppReview.isAvailable()) {
+      final DateTime lastReviewPromptDate = lastReviewPrompt != null ? DateTime
+          .fromMillisecondsSinceEpoch(lastReviewPrompt) : DateTime.now();
+      if (DateTime
+          .now()
+          .difference(lastReviewPromptDate)
+          .inDays > 7) {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  Future<void> _updatePreferences() async {
+    await _sharedPrefs.setInt(InAppReviewConfig.lastReviewPromptDateInMillis,
+        now.millisecondsSinceEpoch,);
+    await _sharedPrefs.setInt(InAppReviewConfig.reviewCountKey,
+      (_sharedPrefs.getInt(InAppReviewConfig.reviewCountKey) ?? 0) + 1,);
+  }
+}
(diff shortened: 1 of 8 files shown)


lib/features/in_app_review/business/in_app_rating_service.dart
```diff
@@ -0,0 +1,59 @@
...
+      InAppReviewConfig.lastReviewPromptDateInMillis,
+    );
+    final reviewCount =
+        _sharedPrefs.getInt(InAppReviewConfig.reviewCountKey) ?? 0;
+    final usageDays = await _getUsageDays();
+
+    if (usageDays > 2 && reviewCount < 3 && await _inAppReview.isAvailable()) {
+      final DateTime lastReviewPromptDate = lastReviewPrompt != null
+          ? DateTime.fromMillisecondsSinceEpoch(lastReviewPrompt)
+          : DateTime.now().subtract(const Duration(days: 8));
```
simon-the-shark, line 41: imo taking 8 days in the past to then pass the if condition feels hacky to me and will be harder to maintain if we'll change something here. Maybe cleaner would be stating the logic in more straightforward way. Example:
```dart
if (lastReviewPrompt == null) {
        return true;
      }
      final DateTime lastReviewPromptDate =
          DateTime.fromMillisecondsSinceEpoch(lastReviewPrompt);
      if (DateTime.now().difference(lastReviewPromptDate).inDays > 7) {
        return true;
      }
```

lib/features/in_app_review/in_app_review.dart
```diff
@@ -0,0 +1,23 @@
...
+import "business/in_app_rating_service_provider.dart";
+
+class InAppReviewWidget extends HookConsumerWidget {
+  const InAppReviewWidget({super.key, required this.child});
+
+  final Widget child;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final inAppRatingService = ref.read(inAppRatingServiceProvider);
```
simon-the-shark, line 16: it's discouraged to ever use `ref.read` in the build method. We should prolly opt for `ref.watch` always to have the newest instance. In this case this shouldn't change at the moment but we don't know what future will bring us.

lib/features/in_app_review/business/in_app_rating_service.dart
```diff
@@ -0,0 +1,59 @@
+import "package:flutter_riverpod/flutter_riverpod.dart";
+import "package:in_app_review/in_app_review.dart";
+import "package:shared_preferences/shared_preferences.dart";
+
+import "../../../config/shared_prefs.dart";
+import "../../../utils/datetime_utils.dart";
+import "../../app_streak/business/get_days_use_case.dart";
+
+class InAppRatingService {
```
simon-the-shark, line 9: btw, very nice service :)

lib/features/in_app_review/business/in_app_rating_service.dart
```diff
@@ -0,0 +1,59 @@
...
+
+  Future<void> requestReviewIfNeeded() async {
+    if (await _canRequestReview()) {
+      await _inAppReview.requestReview();
+      await _updatePreferences();
+    }
+  }
+
+  Future<bool> _canRequestReview() async {
+    final lastReviewPrompt = _sharedPrefs.getInt(
```
simon-the-shark, line 31: so this is purely proposition. I am not fully sure if that's the right way, but you can consider using our wrapper on `DateTime`, called `Timestamp` which would look more or less like this:
```dart
 final lastReviewPrompt = Timestamp.tryParse(
      _sharedPrefs.getString(
        // it uses string instead of int, but this doesn't really matter, don't it?
        InAppReviewConfig
            .lastReviewPromptDateInMillis, // this name prolly should be adjusted then
      ),
    );
```
but again, not sure if this gains us a lot of advantage somewhere

simon-the-shark: Now when I think of it, we're using shared prefs and Timestamp in multiple places in the app already. So we could make an extension on the `SharedPrefs` called `getTimestamp` and `saveTimestamp` that would be reusable and would take the serialization/deserialization in one common place

simon-the-shark (commented):
Very nice and clean piece of code. Good job! 

I have left few propositions :) Maybe you get something usefull from this, but I don't stand strongly for all of them. Tell me what you think.

I'll try to test it in a moment.

lib/features/in_app_review/business/in_app_rating_service.dart
mikolaj-jalocha: Yeah, excellent suggestion. I was trying to keep it simple and eventually ended up even more jumbled  😂
Your next suggestion about making extensions os `SharedPref `object made that part of code even more readable.

lib/features/in_app_review/in_app_review.dart
mikolaj-jalocha: Ofc, my mistake here. I wonder if some linter rule could be configured for this? 🤔

lib/features/in_app_review/business/in_app_rating_service.dart
mikolaj-jalocha: Another good suggestion. Making extension here made code even more readable imo ;)

lib/features/in_app_review/in_app_review.dart
simon-the-shark: yeah, actually it's on my board to set up riverpod linter made by the riverpod author

simon-the-shark (approved):

PR #347, opened 2024-10-26
feat: handle in app deep links

simon-the-shark:
so previously if there was a deeplink in some text or guide or smth we just opened it like any other and relied on the os system to open good tab in our app.

And it even kinda worked. But had some drawbacks:
1. IOS didn't work (xd) - generally in this scenario ios never works (at least that's what i read) so this change is needed here. If we open a link ios assume we can't handle it internally, which is not the case here. we need to handle it ourselves. Disclaimer: i don't think ios deeplinks work at all for our app, but this change removes the worst case where someone can notice it.

2. on android it worked but had one drawback: back button didn't move us back to the place where we opened it. Which imo is kinda bad UX. Now you can always make a nice return to the previous screen.

### DIFF
--- a/lib/config/url_config.dart
+++ b/lib/config/url_config.dart
@@ -1,4 +1,6 @@
 abstract class UrlConfig {
   static const academicCalendarUrl =
       "https://pwr.edu.pl/studenci/kalendarz-akademicki";
+
+  static const topwrUrl = "https://topwr.solvro.pl/";
 }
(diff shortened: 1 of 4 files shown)


mikolaj-jalocha (approved):
LGTM! Fine and simple solution (but I imagine it wasn't that obvious from the start).
BTW deep link handling in our app'd be a great material for an article regarding deep links in complex situations. Assuming we somehow handle iOS's way of thinking 😝

simon-the-shark:
you can even do a solvro talk about it :P - assuming we fix the ios fully

PR #349, opened 2024-10-27
Feat/add riverpod linter

simon-the-shark:
- added riverpod linter.
- we had 13 errors. which of 9 I had resolved
- I had ignored 4 remaining as imo this linter has problems due to a bit complicated abstractions and is issuing this without any reason.

Was planning to do this a long time ago, but custom_lint had a nasty bug

tell me what you think :)


mikolaj-jalocha (approved):
LGTM!

PR #351, opened 2024-10-27
Feat/ animate maps camera

simon-the-shark:
https://github.com/user-attachments/assets/45dfee6d-006c-417a-81fa-c1e7abcf65c0

### DIFF
--- a/lib/features/buildings_view/building_tile.dart
+++ b/lib/features/buildings_view/building_tile.dart
@@ -1,3 +1,5 @@
+import "dart:async";
+
 import "package:flutter/material.dart";
 import "package:flutter_riverpod/flutter_riverpod.dart";
 
@@ -27,7 +29,9 @@ class BuildingTile extends ConsumerWidget {
       subtitle: context.changeNull(building.addressFormatted),
       isActive: isActive,
       onTap: () {
-        ref.read(buildingsMapControllerProvider.notifier).onMarkerTap(building);
+        unawaited(
+          ref.read(buildingsMapControllerProvider).onMarkerTap(building),
+        );
       },
     );
   }
--- a/lib/features/buildings_view/buildings_view.dart
+++ b/lib/features/buildings_view/buildings_view.dart
@@ -1,3 +1,5 @@
+import "dart:async";
+
 import "package:auto_route/auto_route.dart";
 import "package:flutter/material.dart";
 import "package:flutter_map/flutter_map.dart";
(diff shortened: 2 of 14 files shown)


lib/features/buildings_view/building_tile.dart
```diff
@@ -27,7 +29,9 @@ class BuildingTile extends ConsumerWidget {
       subtitle: context.changeNull(building.addressFormatted),
       isActive: isActive,
       onTap: () {
-        ref.read(buildingsMapControllerProvider.notifier).onMarkerTap(building);
+        unawaited(
```
mikolaj-jalocha, line 32: Is this unawaited here because of linter complaints, or sth changed in whole logic that now it's `async`?

lib/features/buildings_view/buildings_view.dart
```diff
@@ -35,7 +37,9 @@ class BuildingsView extends ConsumerWidget {
         point: item.location,
         child: GestureDetector(
           onTap: () {
-            ref.read(buildingsMapControllerProvider.notifier).onMarkerTap(item);
+            unawaited(
```
mikolaj-jalocha, line 40: Same here?

lib/features/map_view/controllers/map_controller.dart
```diff
@@ -1,37 +1,43 @@
...
+class MyMapController<T extends GoogleNavigable> {
+  final Ref ref;
+  final MapControllers<T> mapControllers;
+  MyMapController(this.ref, this.mapControllers);
 
-  @override
-  fl_map.MapController build() {
-    return fl_map.MapController();
+  // nasty, but no other option I think
+  final _controllerCompleter = Completer<AnimatedMapController>();
```
mikolaj-jalocha, line 17: Is this `Completer` here because of build errors? Or could we just create `AnimatedMapController.future`?

mikolaj-jalocha (approved):
LGTM! Nicely worked-around.

Left few questions.

lib/features/buildings_view/building_tile.dart
simon-the-shark: it wasn't async and now it is due to few reasons:
1. we do not have controller straight away, we use `Completer` and await its completion
2. animation also takes time and is async, and previously camera jump wasn't async

lib/features/buildings_view/buildings_view.dart
simon-the-shark: same

lib/features/map_view/controllers/map_controller.dart
simon-the-shark: Completer is just a native dart way to await somewhere until someother part of the code completes with value (in very imperative way with `.complete(value)` - so imo ideal way in this example. But sure, there are few alternative ways to handle this. Riverpod provider could be a solution too, but more bulky

Issue #243, opened 2024-09-27
fix/ fix chart's labels "zawijanie"


simon-the-shark:
![image](https://github.com/user-attachments/assets/78a02bbb-3951-4541-bd40-face677c84ab)

simon-the-shark:
this is not reproducible on all devices

PR #354, opened 2024-10-28
Fix: fix chart labels

mikolaj-jalocha:
#243 

- I couldn't reproduce the bug. Tested on various devices (different width and height of the screen), everything worked fine.  
- I've decided to increase `reservedSize` value from 30 to 32. I think that will resolve imo rare issue.


simon-the-shark (approved):
Yeah, it's not really an important bug to be fixed and seems to be rare. Let's go with it, it shoould be enough

Issue #377, opened 2024-11-03
feat/ add sks people counter


simon-the-shark:
1. fetch csv data from https://live.pwr.edu.pl/sks/sks-data.csv
2. parse csv
3. show number of people in some kind of a button in app bar
4. ![Image](https://github.com/user-attachments/assets/fde880db-adb1-493a-b0e3-27675bc59e96)

simon-the-shark:
possible that we'll get this + trend + parsing on the api side

PR #392, opened 2024-11-10
Feat/add sks people counter

tomasz-trela:
462579375_3855594211427015_7168563900263724153_n
#377 
There is a problem with generating files with linter. So checks doesn't work.


lib/config/env.dart
```diff
@@ -25,4 +25,6 @@ abstract class Env {
   static final String wiredashId = _Env.wiredashId;
   @EnviedField()
   static final String wiredashSecret = _Env.wiredashSecret;
+  @EnviedField()
+  static final String sksUrl = _Env.sksUrl;
```
simon-the-shark, line 29: if you want the build to pass you have to add this to example.env too (even and empty string)

lib/features/sks_people_counter_button/models/sks_user_data.dart
```diff
@@ -0,0 +1,49 @@
...
+      _$SksUserDataFromJson(json);
+}
+
+enum Trend {
+  increasing,
+  decreasing,
+  stable,
+}
+
+class TrendConverter implements JsonConverter<Trend, String> {
```
simon-the-shark, line 28: this wasn't needed. I believe 
```dart
@JsonEnum(fieldRename: FieldRename.screamingSnake)
enum Trend {
  increasing,
  decreasing,
  stable,
}
```
would do the job

lib/config/env.dart
tomasz-trela: ah ok

lib/features/sks_people_counter_button/repository/latest_sks_user_data_repo.dart
```diff
@@ -0,0 +1,14 @@
...
+import "../models/sks_user_data.dart";
+
+part "latest_sks_user_data_repo.g.dart";
+
+@riverpod
+Future<SksUserData> getLatestSksUserData(Ref ref) async {
+  final dio = ref.read(sksClientProvider);
+  final response = await dio.get(SksClientConfig.latestDataEndpoint);
+  return SksUserData.fromJson(response.data as Map<String, dynamic>);
+}
```
simon-the-shark, line 14: for the sake of this task this will be enough, but as a next task (or in this one) we need to add interval refreshing.

Actually I've forced @czaja307 to make a few features for us, so in the response you have 2 fields: `nextUpdateTimestamp` and `isResultRecent`.

The logic we want is:
- if `isResultRecent` == true; then we want to refresh after `nextUpdateTimestamp` passes
- if == false; let's check in a minute (we'll need to see if this behaves as expected but that's the theory at least)

simon-the-shark: but you can delegate it to next task if you prefer

lib/features/sks_people_counter_button/models/sks_user_data.dart
```diff
@@ -0,0 +1,49 @@
+import "package:freezed_annotation/freezed_annotation.dart";
```
simon-the-shark, line 1: TBH, you can disagree with me @tomasz-trela and @mikolaj-jalocha but imo I would rename the feature folder to just `sks_live_people` and place here both the button and the fullscreen chart too - because the model can be reused in the both subfeatures

simon-the-shark: just a quick though, you can disagree - but how about smth like this? what you think guys?
```bash
features/
├─ sks_people_live/
│  ├─ data/
│  │  ├─ models/
│  │  ├─ repositories/
│  ├─ presentation/
│  │  ├─ live_button/
│  │  ├─ chart/
```

tomasz-trela: It's good idea

lib/features/sks_people_counter_button/repository/latest_sks_user_data_repo.dart
tomasz-trela: You can delegate it to next task because I dont have time for next 2 days

simon-the-shark: sure

simon-the-shark (approved):

PR #414, opened 2024-11-19
Add padding to all elements if there is system padding

Rejfi:
By default AppBar adds padding if there is notch or any system padding.
To fix we can add the same padding to others elements and align them in column with logo.


mikolaj-jalocha:
LGTM

PR #417, opened 2024-11-20
Update map_view_config.dart - Use preferred tile.openstreetmap.org URL

danieldegroot2:
See https://github.com/openstreetmap/operations/issues/737


simon-the-shark (approved):
Thanks!

PR #422, opened 2024-11-22
Chore/sks update url and adjust dish model

24bartixx:
Changed sks meals endpoint
In dish model id type change int -> string to match json
If dishes from cache pass empty map

### DIFF
--- a/lib/features/sks-menu/data/repository/sks_menu_repository.dart
+++ b/lib/features/sks-menu/data/repository/sks_menu_repository.dart
@@ -1,5 +1,6 @@
 import "package:collection/collection.dart";
 import "package:fast_immutable_collections/fast_immutable_collections.dart";
+import "package:flutter/foundation.dart";
 import "package:flutter_riverpod/flutter_riverpod.dart";
 import "package:riverpod_annotation/riverpod_annotation.dart";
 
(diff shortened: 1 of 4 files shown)


lib/features/sks-menu/data/repository/sks_menu_repository.dart
```diff
@@ -6,24 +6,21 @@ import "package:riverpod_annotation/riverpod_annotation.dart";
 import "../../../../shared_api_clients/sks_api_client.dart";
 import "../models/dish_category_enum.dart";
 import "../models/sks_menu_data.dart";
+import "../models/sks_menu_response.dart";
 
 part "sks_menu_repository.g.dart";
 
 @riverpod
 Future<IMap<DishCategory, List<SksMenuDish>>> getSksMenuData(Ref ref) async {
+  const mealsEndpoint = "/api/v1/meals/current";
```
mikolaj-jalocha, line 15: In my `.env` file SKS_URL is as follows: `SKS_URL="https://sks-api.topwr.solvro.pl/api/v1"` in other parts of code  endpoints are used that adapt to this format. 
So endpoint here should start with `/meals..`

mikolaj-jalocha (changes requested):
Good work here! I've spotted only one change here.

@simon-the-shark  pls take a look as well.

lib/features/sks-menu/data/repository/sks_menu_repository.dart
24bartixx: Okay

lib/features/sks-menu/presentation/sks_menu_screen.dart
```diff
@@ -52,6 +52,8 @@ class _SksMenuView extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    debugPrint("_SksMenuView");
```
simon-the-shark, line 55: it's good you've used debugPrint instead of print, but afaik we don't really need it, so it can be deleted

24bartixx:
Thank You @simon-the-shark but I know it's nothing special still.
Nevertheless, I think it's done
Waiting for review

simon-the-shark (approved):
LGTM 😎😎🔥

PR #426, opened 2024-11-23
Feat/caching framework

tomasz-trela:
Cache manager is implemented as singleton because it's not allowed to create 2 cache managers with the same key.
#382


lib/api_base_rest/cache/cache.dart
```diff
@@ -0,0 +1,35 @@
...
+import "dart:typed_data";
+
+import "package:flutter_cache_manager/flutter_cache_manager.dart";
+import "package:flutter_riverpod/flutter_riverpod.dart";
+import "../../shared_api_clients/sks_api_client.dart";
+import "cache_manager.dart";
+
+Future<T> getAndCacheData<T>(
+  String url,
+  Ref ref,
```
simon-the-shark, line 11: did you consider making it an extension on `Ref` instead of a normal function?

lib/api_base_rest/cache/cache.dart
```diff
@@ -0,0 +1,35 @@
...
+  CacheManager cacheManager,
+  T Function(Map<String, dynamic> json) fromJson,
+  bool Function() condition,
```
simon-the-shark, line 14: this name can prolly be better than `condition`

lib/api_base_rest/cache/cache.dart
```diff
@@ -0,0 +1,35 @@
...
+) async {
+  final cachedFile = await cacheManager.getFileFromCache(url);
+  if (cachedFile != null && condition()) {
+    final cachedData = await cachedFile.file.readAsString();
+    final data = fromJson(
+      jsonDecode(cachedData) as Map<String, dynamic>,
+    );
+    return data;
+  }
+  final dio = ref.read(sksClientProvider);
```
simon-the-shark, line 24: we can prolly accept Dio as argument too to make it reusable for other URLs too.

simon-the-shark: Or actually we can maybe make one Dio singleton and always accept full url instead?

simon-the-shark: what you think is better

lib/api_base_rest/cache/cache_manager.dart
```diff
@@ -0,0 +1,22 @@
+import "package:flutter_cache_manager/flutter_cache_manager.dart";
+import "package:flutter_riverpod/flutter_riverpod.dart";
+import "package:riverpod_annotation/riverpod_annotation.dart";
+
+part "cache_manager.g.dart";
+
+@Riverpod(keepAlive: true)
+CacheManager cacheManagerSingleton(Ref ref) {
```
simon-the-shark, line 8: since you're using this singleton anyway, you can inject this into `getAndCacheData` with  `ref.watch` and not as a function argument

lib/api_base_rest/cache/cache.dart
simon-the-shark: also `ref.watch` is prolly more proper in this situation than `ref.read`

lib/api_base_rest/cache/cache_manager.dart
```diff
@@ -0,0 +1,22 @@
...
+  return CacheManager(
+    CacheManagerConfig.jsonCacheManagerConfig,
+  );
+}
+
+abstract class CacheManagerConfig {
+  static final jsonCacheManagerConfig = Config(
+    "JsonCacheManager7Days",
+    stalePeriod: const Duration(days: 7),
```
simon-the-shark, line 17: so I'm thinking how can we use different TTL durations for different endpoints

simon-the-shark: So I'm thinking either separate cache managers for every endpoint or separate cache managers for different TTL durations

simon-the-shark: e.g. maybe smth like this: 
```dart
@Riverpod(keepAlive: true)
CacheManager restCacheManage(Ref ref, int ttlDays) {
  return CacheManager(
    Config(
      "JsonCacheManager${ttlDays}Days",
      stalePeriod: Duration(days: ttlDays),
      maxNrOfCacheObjects: 10,
    ),
  );
}
```

lib/api_base_rest/cache/cache.dart
simon-the-shark: I guess we can migrate to using single Dio client and accept full urls and then the functoin could have signuture like this (assuming its an extension on ref)

```dart
getAndCacheData(
  String url,
  int ttlDays,
  T Function(Map<String, dynamic> json) fromJson,
  bool Function() extraValidityCheck,
)
```

simon-the-shark (commented):
Nice job Tomuś. Most of my comments is just thinking how can we do a little refactor here to make it reusable also for other backend services, not only sks.

A lot of room for discussion for sure.

@mikolaj-jalocha can look too.

My only concern is if saving jsons as files won't be slower than using single NoSQL database? But hmm, the package is pretty easy to use so we can leave it this way and migrate in case of any performance issues.

lib/api_base_rest/cache/cache.dart
tomasz-trela: Yes that's good Idea I think

tomasz-trela:
I wanted to try this package first because it is very easy to implement and use it.
Tomorrow I will fix these issues

simon-the-shark (approved):
LGTM😎😎 Good job Tomuś. 

I'll try this out in some repo (maybe) today.

PR #428, opened 2024-11-24
Feat/add lottie animation

mmzek:
lottie-animation
lottie-animation1

When error occurres or SKS's closed, the lottie animation "closed" is showing on SKS Menu.
In my opinion, the offline menu functionality is out of scope for this task and should be addressed as a separate task.


mikolaj-jalocha:
It appears that again some weird linter error occurred. Imo this happens only if build_runner has not been sent before push. 
Linter complaints should disapear once merged into main.

android/app/build.gradle
```diff
@@ -35,8 +35,8 @@ android {
     ndkVersion flutter.ndkVersion
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
```
tomasz-trela, line 39: Why did you change it?

android/gradle.properties
```diff
@@ -1,3 +1,5 @@
 org.gradle.jvmargs=-Xmx4G
 android.useAndroidX=true
 android.enableJetifier=true
+org.gradle.java.home=C:/Program Files/Eclipse Adoptium/jdk-17.0.13.11-hotspot
```
tomasz-trela, line 4: image
why did you add this java path?

tomasz-trela (commented):
I don't think it's good to make these changes (gradle properties and java version)

android/app/build.gradle
mikolaj-jalocha: Probably this was needed  because of some demands of Android Studio. My advice is to make such changes only locally (at least for now, we can discuss global change of java version ) and don't push them on github. 

Please change them back to 1.8

android/gradle.properties
mikolaj-jalocha: Same here,  we can ommit this file when pushing chagnes on remote. 

Please change it back

assets/animations/closed.json
mikolaj-jalocha, line 1: I like the anim :-)

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,70 @@
+import "package:flutter/material.dart";
+import "package:lottie/lottie.dart";
+
+import "../../../../../theme/app_theme.dart";
+import "../../../../../utils/context_extensions.dart";
+import "../../../../gen/assets.gen.dart";
+
+class LottieLoading extends StatefulWidget {
```
mikolaj-jalocha, line 8: I'd think about better naming, that reflects where such lottie animation is used, eg. "SKSMenuLottieAnimation" or sth like that.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,70 @@
...
+  });
+
+  @override
+  _LottieLoadingState createState() => _LottieLoadingState();
+}
+
+class _LottieLoadingState extends State<LottieLoading> {
+  final bool _showOldMenu = false;
+
+  void _gotoOldMenu() {
```
mikolaj-jalocha, line 20: Some debugging tools i suposse :-) Should be deleted

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,70 @@
...
+class LottieLoading extends StatefulWidget {
+  const LottieLoading({
+    super.key,
+  });
+
+  @override
+  _LottieLoadingState createState() => _LottieLoadingState();
+}
+
+class _LottieLoadingState extends State<LottieLoading> {
```
mikolaj-jalocha, line 17: Actually, I think that we don;t need here state class.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,70 @@
...
+    return Scaffold(
+      body: Column(
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        children: [
+          const Spacer(),
+          Center(
+            child: SizedBox(
+              width: 200,
+              height: 200,
+              child: Lottie.asset(
```
mikolaj-jalocha, line 38: Lottie here could be a little more customized. Ie. cache, frame rate. Please see reference in file used somewhere else: 

image

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,70 @@
...
+                Assets.animations.closed,
+                repeat: false,
+              ),
+            ),
+          ),
+          const Spacer(),
+          Align(
```
mikolaj-jalocha, line 45: As this tasks's scope was reduced, we don;t need such a button here. 

But it would be great if some message's displayed when error occured. Such us, menu is not available, pls try again later, or sth like that. So that user could understand what's going on

pubspec.yaml
```diff
@@ -149,6 +149,7 @@ flutter:
     - assets/animations/search.json
     - assets/animations/offline.json
     - assets/animations/bug.json
+    - assets/animations/closed.json
```
mikolaj-jalocha, line 152: WDYT about a little more descriptive name: sks_closed.json ?

mikolaj-jalocha (changes requested):
First of all, congrats of your first commit here and working code!  Animation looks great😎

I left few suggestions of mine regarding UI. I think button could be removed (as of now) and some text regarding error could be displayed as well. 

I also left notes regarding better use of lottie and some regarding changes of files that should only be made locally.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,70 @@
...
+  Widget build(BuildContext context) {
+    return Scaffold(
+      body: Column(
+        mainAxisAlignment: MainAxisAlignment.spaceBetween,
+        children: [
+          const Spacer(),
+          Center(
+            child: SizedBox(
+              width: 200,
+              height: 200,
```
tomasz-trela, line 37: Some values can be separated to the ui_config as costant values.

tomasz-trela (changes requested):
We don't really want to have magic nubers in code.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
simon-the-shark: Actually I would reccomend using `SizedBox.square` constructor which removes duplication and config file extraction can prolly be skipped then.

simon-the-shark (commented):
So good job and congrats for the first completed task.

I've left one note about the sized box, but everything else was pretty much covered by Tomek and Mikołaj, so good job guys too.

As for java build settings, let's not change anything globally for now. I've actually seen that Andrea has made some fancy tool related on the topic, so someone can investigate on that: https://gist.github.com/bizz84/605e2ca2088cb4acb7a076ca993f41cd?trk=public_post_comment-text But let's rollback for now

mmzek:
Sorry for pushing java path etc., @mikolaj-jalocha was right (some demands of Android Studio), will not happen again. 
I applied changes you mentioned in review and added text describing whats wrong on the screen. Still have some problems with linter even tho flutter analyze shows no issues found.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
mikolaj-jalocha, line 1: Name of the file could be different since this animation's not responsible for loading. In fact we could use that class inside top level sks_menu_screen as a private class.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,41 @@
...
+              child: Lottie.asset(
+                Assets.animations.sksClosed,
+                fit: BoxFit.cover,
+                repeat: false,
+                frameRate: const FrameRate(LottieAnimationConfig.frameRate),
+                renderCache: RenderCache.drawingCommands,
+              ),
+            ),
+          Align(
+              child: Text(
```
mikolaj-jalocha, line 31: Maybe it'd be good to add  detailed error message (error message that comes from api). Maybe in small print and in different color.

lib/features/sks-menu/presentation/widgets/sks_menu_loading.dart
```diff
@@ -0,0 +1,41 @@
...
+import "../../../../config/ui_config.dart";
+import "../../../../gen/assets.gen.dart";
+
+class SKSMenuLottieAnimation extends StatelessWidget {
+  const SKSMenuLottieAnimation({
+    super.key,
+  });
+
+  @override
+  Widget build(BuildContext context) {
```
mikolaj-jalocha, line 15: It'd be great to add logger somewhere here:

`Logger().e(error.toString());`

lib/features/sks-menu/presentation/sks_menu_screen.dart
```diff
@@ -63,11 +56,7 @@ class _SksMenuView extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     if (sksMenuData.meals.isEmpty) {
-      return const Scaffold(
-        body: Center(
-          child: CircularProgressIndicator(),
-        ),
-      );
+      return const SKSMenuLottieAnimation();
```
mikolaj-jalocha, line 59: Here we should check if menu's online or not and then show animation (list of meals now is only empty in case of error, if sks's closed cached data are provided, so this condition'll never work)

mikolaj-jalocha (changes requested):
Good job! Code looks cleaner now. I tested it and animations works good on my device.
I left few notes (one is crucial, pls see what happens if menu is offline)

lib/features/sks-menu/presentation/sks_menu_screen.dart
```diff
@@ -63,11 +56,7 @@ class _SksMenuView extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
     if (sksMenuData.meals.isEmpty) {
```
mikolaj-jalocha, line 58: Here we should check if menu's online or not and then show animation (list of meals now is only empty in case of error, if sks's closed cached data are provided, so this condition'll never work)

mikolaj-jalocha (changes requested):
and one more

mikolaj-jalocha (dismissed):
LGTM! 🚀

simon-the-shark (approved):
Looks good to me. I would maybe think of changing the animation color, but that's secondary consern. Merging it now. Good job!

PR #436, opened 2024-11-27
feat: implement review request open store

thesun901:
navigation_tab_view.dart: add revirew openStoreListing behavior to review button

pretty much one liner.
Button opens store to rate app - according to guidlines

requestReview() (avoid)
image

openStoreListing() guidelines
image

I am open to disscusion

### DIFF
--- a/lib/features/navigation_tab_view/navigation_tab_view.dart
+++ b/lib/features/navigation_tab_view/navigation_tab_view.dart
@@ -1,6 +1,7 @@
 import "package:auto_route/auto_route.dart";
 import "package:flutter/material.dart";
 import "package:flutter_riverpod/flutter_riverpod.dart";
+import "package:in_app_review/in_app_review.dart";
 
 import "../../config/ui_config.dart";
 import "../../theme/app_theme.dart";
@@ -36,8 +37,8 @@ class NavigationTabView extends ConsumerWidget {
           ),
         ),
         child2: SmallTileCard(
-          onTap: () {
-            // TODO(simon-the-shark): implement review request
+          onTap: () async {
+            await InAppReview.instance.openStoreListing();
           },
           title: context.localize.leave_a_review,
           icon: Icon(


lib/features/navigation_tab_view/navigation_tab_view.dart
```diff
@@ -36,8 +37,8 @@ class NavigationTabView extends ConsumerWidget {
           ),
         ),
         child2: SmallTileCard(
-          onTap: () {
-            // TODO(simon-the-shark): implement review request
+          onTap: () async {
+            await InAppReview.instance.openStoreListing();
```
mikolaj-jalocha, line 41: Overall looks, I didn't run it but I assume it works (you can provide some picture if that's not a problem)

I've got one thing, 
WDYT about creating such a method in `InAppRatingService` class and call it from `navigation_tab_view`? Looks like overkill but imagine if we (for some reasons) would have to change library for in app review? @simon-the-shark  @tomasz-trela wdyt?

mikolaj-jalocha (commented):
Please, see the comment and let's w8 for opinions of others

lib/features/navigation_tab_view/navigation_tab_view.dart
thesun901: some images

before click:
image
after click:
image


didnt check that on ios tho :(

if it comes to making specified class I will wait for opinions of others and I can handle implementation if others agree on doing that!

tomasz-trela: > Overall looks, I didn't run it but I assume it works (you can provide some picture if that's not a problem)
> 
> I've got one thing, 
> WDYT about creating such a method in `InAppRatingService` class and call it from `navigation_tab_view`? Looks like overkill but imagine if we (for some reasons) would have to change library for in app review? @simon-the-shark  @tomasz-trela wdyt?  


I think it is a good practice.

simon-the-shark: And I think you're guys overcomplicating oneliner solution 😭😭😭

mikolaj-jalocha: Since've got that service already defined, it's almost no cost defining method in there. From `navigation_tab_view` perspective nothing changes and code looks cleaner, we can controll anything from one place

simon-the-shark (approved):
LGTM.

Separate service is a bit too much for a one liner.

We could do global util at most, but at the moment it's the only place in the app where we call it so it's good to go. In the future if we do this at few places in the codebase, then a global util could be a nice refactor.

Merging it now.

simon-the-shark:
Or actually since the Service is alresdy in place, this isnt that bad of an idea @thesun901 @mikolaj-jalocha

Issue #444, opened 2024-11-29
fix/fix-padding-at-last-tab


tomasz-trela:

tomasz-trela:
![Image](https://github.com/user-attachments/assets/fd9630e6-ee67-4a96-81e3-ca9fd58ee9eb)

PR #445, opened 2024-11-29
refactor: fix padding

tomasz-trela:
Simulator Screenshot - iPhone 16 Plus - 2024-11-29 at 20 15 32
image
#444

### DIFF
--- a/lib/config/ui_config.dart
+++ b/lib/config/ui_config.dart
@@ -195,7 +195,7 @@ abstract class SksConfig {
   static const sizedBoxWidth = 5.0;
   static const radius = 8.0;
   static const innerPadding = EdgeInsets.symmetric(horizontal: 8, vertical: 4);
-  static const outerPadding = EdgeInsets.only(right: 12, bottom: 2);
+  static const outerPadding = EdgeInsets.only(right: 24, bottom: 2);
 }
 
 abstract class NavigationTabViewConfig {
--- a/lib/config/ui_config.dart
+++ b/lib/config/ui_config.dart
@@ -195,7 +195,8 @@ abstract class SksConfig {
   static const sizedBoxWidth = 5.0;
   static const radius = 8.0;
   static const innerPadding = EdgeInsets.symmetric(horizontal: 8, vertical: 4);
-  static const outerPadding = EdgeInsets.only(right: 24, bottom: 2);
+  static const outerPaddingLarge = EdgeInsets.only(right: 24, bottom: 2);
+  static const outerPadding = EdgeInsets.only(right: 12, bottom: 2);
 }
 
 abstract class NavigationTabViewConfig {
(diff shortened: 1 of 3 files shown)


lib/config/ui_config.dart
```diff
@@ -195,7 +195,7 @@ abstract class SksConfig {
   static const sizedBoxWidth = 5.0;
   static const radius = 8.0;
   static const innerPadding = EdgeInsets.symmetric(horizontal: 8, vertical: 4);
-  static const outerPadding = EdgeInsets.only(right: 12, bottom: 2);
+  static const outerPadding = EdgeInsets.only(right: 24, bottom: 2);
```
simon-the-shark, line 198: did you check that this padding isn't used in any other place?

tomasz-trela: Yes it is used in sks menu tab in SksUserButton. But I can't checkout it because sks menu is offline
Simulator Screenshot - iPhone 16 Plus - 2024-11-29 at 20 43 03

tomasz-trela:
So we have to wait till Monday :/

lib/config/ui_config.dart
mikolaj-jalocha: try to comment condition that checks if menu is offline. We always have some kind of data (even old).

 ```dart
 if (!sksMenuData.isMenuOnline) {
      return const _SKSMenuLottieAnimation();
    }
```
line 61 of `sks_menu_screen.dart`

mikolaj-jalocha (commented):
Please check if paddings didn't disturb other parts of UI. I left one comment regarding how to view sks's data even if menu's offline.

tomasz-trela:
But I've got an exception 
Zrzut ekranu 2024-11-30 o 15 47 36
So commenting this code gives no result

mikolaj-jalocha:
> But I've got an exception Zrzut ekranu 2024-11-30 o 15 47 36 So commenting this code gives no result

that's weird. I'll check on that

mikolaj-jalocha:
> But I've got an exception Zrzut ekranu 2024-11-30 o 15 47 36 So commenting this code gives no result

the oldest joke in the industry, but work on my computer XD 

image

 it's the only thing that I've done.

tomasz-trela:
Simulator Screenshot - iPhone 16 Plus - 2024-11-30 at 20 33 31
So this screen looks like this and I would do separate value of padding these screen.

simon-the-shark:
> Simulator Screenshot - iPhone 16 Plus - 2024-11-30 at 20 33 31
> So this screen looks like this and I would do separate value of padding these screen.

Yes, lets separate it

tomasz-trela:
Simulator Screenshot - iPhone 16 Plus - 2024-11-30 at 20 43 14
Now it looks like it was previously

simon-the-shark (approved):

PR #446, opened 2024-11-30
fix/ fix shadows on filter screen

BombardierBulge:
- Added padding to button so shadow won't get clipped 
image


mikolaj-jalocha (dismissed):
Tbh this bug was extremely hard to spot XD  But according to your screen shot everything works as expected. Please see my comment and think if you want to change, let me know.

simon-the-shark (commented):
Yeah, good job here. When you correct this padding that Mikołaj noted, please ping me before merging. I need to make a rebase to fix some commit history first.

PR #448, opened 2024-11-30
418 change searchs results order at buildings screen

jakubbaldyga:
**Repaired the issue.**
The behavior for searching buildings is now as follows:
 - No input: All buildings are displayed.
 - One-letter input: A check is performed to see if it matches a building code. If it does, buildings are filtered by the building code. If not, they are filtered by name and address.
- Multiple letters: Buildings are filtered by code, name, and address.

### DIFF
--- a/lib/features/buildings_view/model/building_model.dart
+++ b/lib/features/buildings_view/model/building_model.dart
@@ -1,9 +1,12 @@
 import "package:latlong2/latlong.dart";
 
+import "../../../config/map_view_config.dart";
 import "../../map_view/controllers/controllers_set.dart";
 import "../repository/buildings_repository.dart";
 
 class BuildingModel extends Building implements GoogleNavigable {
+  static Set<String> buildingCodesLowerCase = <String>{};
+
   BuildingModel.from(Building building)
       : super(
           id: building.id,
@@ -14,7 +17,11 @@ class BuildingModel extends Building implements GoogleNavigable {
           cover: building.cover,
           naturalName: building.naturalName,
           disableBuildingPrefix: building.disableBuildingPrefix,
-        );
+        ) {
+      if(building.name[1] == BuildingSearchConfig.buildingCodeSeperator) {
+        buildingCodesLowerCase.add(building.name[0].toLowerCase());
+      }
+  }
 
   @override
   LatLng get location => LatLng(latitude, longitude);
--- a/lib/features/buildings_view/model/building_model.dart
+++ b/lib/features/buildings_view/model/building_model.dart
@@ -1,11 +1,13 @@
 import "package:latlong2/latlong.dart";
 
 import "../../../config/map_view_config.dart";
+import "../../../utils/contains_number.dart";
 import "../../map_view/controllers/controllers_set.dart";
 import "../repository/buildings_repository.dart";
 
 class BuildingModel extends Building implements GoogleNavigable {
-  static Set<String> buildingCodesLowerCase = <String>{};
+  static final BuildingCodesLowerCase _buildingCodesLowerCase =
+      BuildingCodesLowerCase();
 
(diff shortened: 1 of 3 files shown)


tomasz-trela (approved):
Everything works as expected! Code looks good. Let's wait for @mikolaj-jalocha opinion.

lib/features/buildings_view/model/building_model.dart
```diff
@@ -14,11 +19,46 @@ class BuildingModel extends Building implements GoogleNavigable {
...
   String? get addressFormatted =>
       addres?.replaceFirst(",", "\n").replaceAll("\n ", "\n");
+
+  static bool isBuildingCode(String buildingCode) {
+    return _buildingCodesLowerCase.contains(buildingCode) ||
+        buildingCode.containsNumber();
+  }
+}
+
+class BuildingCodesLowerCase {
```
simon-the-shark, line 38: Sorry to hurt your feelings, but I'm not a biggest fan of this approach 😭😭 I would highly recomend doing this in some functional programming manner instead. 

The logic itself looks fine, I'm just not a biggest fan of this static class. This could be much cleaner and shorter using more declarative and less imperative paradigm. Maybe we could use the benefits of riverpod here too?

I can give you some example code in a second

simon-the-shark:
Okey, so first of all: this is only my proposition. We can discuss it, but it generally shows the direction we should go (at least in my opinion). I'm open for discussion. Also I didn't test it very well. 

1. `parsedBuildingCode` is fine. I would maybe move it to `BuildingModel` as `String? get parsedBuildingCode { ... }`, because we're deleting the class it is in now
2. I would delete the `BuildingCodesLowerCase` class, cause it's very imperative approach. We can do similar logic with riverpod and functional paradigm:
```dart
part "all_buildings_lower_case.g.dart"; // not sure where we wanna place this file and how to call it

@riverpod
Future<ISet<String>> allBuildingsCodesLowerCase(Ref ref) async {
  final buildings = await ref.watch(buildingsRepositoryProvider.future);
  return buildings
      .map((e) => e.parsedBuildingCode?.toLowerCase())
      .whereType<String>()
      .toISet();
}

extension IsStringABuildingCodeX on Ref {
  bool isStringABuildingCode(String potencialBuildingCode) {
    final allLowerCaseBuildingsCodes =
        watch(allBuildingsCodesLowerCaseProvider).requireValue;
    return allLowerCaseBuildingsCodes
            .contains(potencialBuildingCode.toLowerCase()) ||
        potencialBuildingCode.containsNumber();
  }
}
```

3. Now we can use `ref.isStringABuildingCode(some str)` to check the logic we previously checked using buildings model method. The `filter` implementation is also really nice.
```dart
// ...
case 1:
        if (ref.isStringABuildingCode(filterStr)) {
//...
```


@mikolaj-jalocha  @tomasz-trela @jakubbaldyga  tell me what you think. Prolly small refactor can even more simplify this code or make it cleaner, but it's more or less how I would do it.

jakubbaldyga:
I love your idea, I will implement it today and let you know how it went😄

jakubbaldyga:
@simon-the-shark  I copy-pasted your code, and it worked as expected👍. I created the utils folder like it was done in other features folders. And I moved utils.dart and created building_codes_utils.dart with those functions. Thanks for your time!

mikolaj-jalocha:
> @simon-the-shark I copy-pasted your code, and it worked as expected👍. I created the utils folder like it was done in other features folders. And I moved utils.dart and created building_codes_utils.dart with those functions. Thanks for your time!

Please mark comments that your PR fix includes as resolved ;)

mikolaj-jalocha (approved):
lgtm

simon-the-shark (approved):

Issue #378, opened 2024-11-03
feat/ add sks chart


simon-the-shark:
- this uses shuld reuse fetching data from the task: https://github.com/Solvro/mobile-topwr/issues/377
- but the parsing might be more specific for this case
- this should be a full-screen closable chart. a bit like the original site: https://live.pwr.edu.pl/sks/
- My very ruff visualization: ![Image](https://github.com/user-attachments/assets/21bade36-d76f-403b-9f91-790e34a819ce)
- but basically this can possily reuse the chart from parkings view? - if will look fine
- basically consult the team when you have first example UI in flutter

simon-the-shark:
when you wait for the #377 to complete, you can start crafthing the ui on mocked data

simon-the-shark:
possible that we'll get this + trend + parsing on the api side

czaja307:
`https://sks-api.topwr.solvro.pl/api/v1/sks-users/today`

PR #458, opened 2024-11-30
SKS chart

mikolaj-jalocha:
- for clarity the chart shows only 6 periods of the given time, 3 in the past, 3 in the future
- didn't test in on high numbers like 100, but should scale well
- any thing else we can add here? Maybe some note about from where the data comes from? Btw wdyt about making an article about sks's chart data on our blog? How it started, who's behind etc. as a fun fact :-) We could then link to that from our app. 
chart
#378


simon-the-shark:
Hey! Didn't look more into this, but I have some quick thoughts/suggestions about the design:
1. vertical scale labels maybe should be on the left side - like in parking chart to be consistent??
2. I think we need at least 4h scope of time for it to be useful
3. We can have few times more bars - just the labels don't need to be for every one of them. I think when we have more dense bars, this should look better. Maybe more like in the Google Maps? I thinks it's pretty neatly done there.
4. maybe some small and vague horizontal lines?? - quite optional
5. maybe smaller corner radius of the bars??????????
6. maybe let's put the whole chart into some greyish container? - to be more consistent with the parkings?


not about the chart: maybe we should also show current people counter and trend in the bottom drawer too? cause the app bar one is kinda in the background. In the parkings we have the current number too, so it would be more consistent.


Tell me what you think! Hopefully I don't interfere with your vision too much 🦔

tomasz-trela:
I think vertical bars should be at right site. Similar to parkings. I think interval should be bigger.

tomasz-trela:
> Hey! Didn't look more into this, but I have some quick thoughts/suggestions about the design:
> 
> 1. vertical scale labels maybe should be on the left side - like in parking chart to be consistent??
> 2. I think we need at least 4h scope of time for it to be useful
> 3. We can have few times more bars - just the labels don't need to be for every one of them. I think when we have more dense bars, this should look better. Maybe more like in the Google Maps? I thinks it's pretty neatly done there.
> 4. maybe some small and vague horizontal lines?? - quite optional
> 5. maybe smaller corner radius of the bars??????????
> 6. maybe let's put the whole chart into some greyish container? - to be more consistent with the parkings?
> 
> not about the chart: maybe we should also show current people counter and trend in the bottom drawer too? cause the app bar one is kinda in the background. In the parkings we have the current number too, so it would be more consistent.
> 
> Tell me what you think! Hopefully I don't interfere with your vision too much 🦔

Oh I didn't see your message.

mikolaj-jalocha:
Extremely fast visualization of your propositions:

- imo gray background doesn't look good
- certain hours (ie. each 3rd or even 4th) could be placed obliquely to the graph)
- live counter is a good option, I will implement it
- smaller corner radius's good idea
- Do you think that left axis need title? 
- with 4 hours of scope (5 min intervals) chart is not readable. In following scenario prob. line chart'd be better. 


chart

tomasz-trela:
> Extremely fast visualization of your propositions:
> 
> * imo gray background doesn't look good
> * certain hours (ie. each 3rd or even 4th) could be placed obliquely to the graph)
> * live counter is a good option, I will implement it
> * smaller corner radius's good idea
> * Do you think that left axis need title?
> * with 4 hours of scope (5 min intervals) chart is not readable. In following scenario prob. line chart'd be better.
> 
> chart

I think grey background looks bad, we don't need left title. Maybe change intervals to 10 minutes or try with line chart.

24bartixx:
Greate job!
I suggest adding orange opacity without a border to the prediction bars rather than outlined bars for better readability 😄

simon-the-shark:
Eventually the idea makes sense, but probably we could wait until we make the events/calendar feature that is planned next year.

simon-the-shark:
And detecting "unusual activity" seems like a backend task

simon-the-shark:
We can drop it to placeholder ideas for now?

mikolaj-jalocha:
> And detecting "unusual activity" seems like a backend task

Maybe in ultra easy version it could be as simple as displaying widget when measured activity was higher that forecasted one

mikolaj-jalocha:
WDYT?

24bartixx:
I would also consider changing the blue color text at the top and on the bottom to a black one for my optical satisfaction as it seems more natural to me. But it's only my preference and it's optional.

mikolaj-jalocha:
Zrzut ekranu 2024-12-16 o 00 02 18

simon-the-shark:
> I would also consider changing the blue color text at the top and on the bottom to a black one for my optical satisfaction as it seems more natural and minimalistic to me. But it's only my preference and it's optional.

yeah, imo the way to go. we have already too much color in this view

Issue #421, opened 2024-11-21
feat/ add info dialog about notifications


simon-the-shark:
When tapped on the button, show infomartive dialog that "We're not using push notifications at the moment, but we're planning to and you can aggree to them now" and after user presses ok it should be prompted for permissions

showed up after #416 

![Image](https://github.com/user-attachments/assets/2f53f209-6371-470a-9acb-fdc06615c3a8)

PR #460, opened 2024-12-01
Feat: Add info dialog about notifications

gry-mar:
Created custom Alert Dialog that can be reused for the app version and push notification alerts depending on specified Dialog Content. 

#421


lib/widgets/my_alert_dialog.dart
```diff
@@ -0,0 +1,91 @@
...
+    required this.onConfirmTapped,
+    required this.confirmText,
+  });
+
+  @override
+  Widget build(BuildContext context) {
+    return AlertDialog(
+      content: Padding(
+        padding: const EdgeInsets.symmetric(
+          horizontal: AboutUsConfig.dialogHorizontalPadding,
```
mikolaj-jalocha, line 53: Since dialog's now global,  wdyt about moving that to separate config file?

mikolaj-jalocha (approved):
Great work here! I've nothing to add (besides localization of padding's values). LGTM 🚀🚀🚀🚀

lib/widgets/my_alert_dialog.dart
gry-mar: Yes, you're right! So I'll move it to new config class like `AlertDialogConfig`

mikolaj-jalocha (approved):

PR #461, opened 2024-12-01
Feat/see latest available menu

mmzek:
When sks's closed I added a button for a user to see menu from the latest update. I made a text and the button appear after the lottie animation ends.  
button1
button


lib/features/sks-menu/data/repository/sks_menu_repository.dart
```diff
@@ -7,6 +7,30 @@ import "../models/sks_menu_response.dart";
 
 part "sks_menu_repository.g.dart";
 
+@riverpod
```
simon-the-shark, line 10: It's nice to see you've learnt some riverpod! It's very nice to see. Hovewer, riverpod is best about handling global/business state and here it's more about local/emphermal state and in our app we try to use Flutter Hooks for local state scenarios. We still have some old part of the code that use rivepod for semi-global state, but even so then it needs to be scoped to the route. And scoping riverpod is kinda nasty. I would reccommend using hooks for this scenario. 

https://riverpod.dev/docs/essentials/do_dont#avoid-using-providers-for-ephemeral-state: 

Zrzut ekranu 2024-12-2 o 14 37 10



PS. even if we stay with the riverpod, it needs to be scoped to the root and `repository` is not a proper file for this. Since it's UI related state, it belongs to presentation layer. In case of riverpod you could make `something_controller.dart` file - or just use flutter hooks. They should be pretty easy to implement ;)

lib/features/home_view/widgets/parkings_section.dart
```diff
@@ -25,6 +25,10 @@ class ParkingsSection extends ConsumerWidget {
             actionTitle: context.localize.map_button,
             onClick: ref.navigateParkings,
           ),
+          FilledButton(
+            onPressed: ref.navigateToSksMenu,
+            child: const Text("navigate to sks menu"),
```
simon-the-shark, line 30: This should prolly be localized string.

simon-the-shark (commented):
@mikolaj-jalocha @tomasz-trela I didn't look through the code in very much detail. But I have one big note about the local vs global state and recommend using hooks for this scenario. This is also meant to be a chapter in our Flutter Handbook and is mentioned in the riverpod docs and in the "Flutter Engineering" book that we have on out Google Drive.

You can guys make a more detailed review if you see something interesting.

mikolaj-jalocha (changes requested):
As a Szymon said, please convert existing solution to one based on Flutter Hooks and then re-request review. 
In case of problems, hmu ;)

simon-the-shark:
What's the status here?

lib/features/sks-menu/presentation/sks_menu_screen.dart
```diff
@@ -30,17 +32,27 @@ class SksMenuView extends ConsumerWidget {
...
+        if (!sksMenuData.isMenuOnline && !isLastMenuButtonClicked.value) {
+          return _SKSMenuLottieAnimation(
+            onShowLastMenuTap: () {
+              isLastMenuButtonClicked.value = true;
+            },
+          );
+        }
+        if (sksMenuData.isMenuOnline && isLastMenuButtonClicked.value) {
+          isLastMenuButtonClicked.value = false;
+        }
```
simon-the-shark, line 49: so we probably shouldn't change state during build method itself. For this we place it inside `useEffect` hook for example? Or just adjust the logic for this to be not necessary.

simon-the-shark: So if you wanna change this, then this useEffect hook could probably be a better idea:
```dart
final asyncSksMenuData = ref.watch(getSksMenuDataProvider);
    final isLastMenuButtonClicked = useState(false);
    useEffect(() {
      if ((asyncSksMenuData.value?.isMenuOnline ?? false) &&
          isLastMenuButtonClicked.value) {
        isLastMenuButtonClicked.value = false;
      }
      return null;
    }, [
      asyncSksMenuData.value?.isMenuOnline,
      isLastMenuButtonClicked.value,
    ]);
```
But it's long and nasty.

But if you look down to you other widget, you don't show the animation if the menu is online anyway - so imo you can skip it. I know you wanted to make sure that we cover the edge case that menu is offline, someone clicked, it then went offline and online again and for us this would mean it's clicked - but imo that's not a problem. Apart from this case, this if statement didn't change anything anyway. So imo you can just skip it.

lib/features/sks-menu/presentation/sks_menu_screen.dart
```diff
@@ -52,13 +64,19 @@ class SksMenuView extends ConsumerWidget {
...
+  });
 
   final SksMenuResponse sksMenuData;
   final String? appBarPopTitle;
+  final bool isLastMenuButtonClicked;
+
   @override
   Widget build(BuildContext context) {
-    if (!sksMenuData.isMenuOnline) {
+    if (!isLastMenuButtonClicked && !sksMenuData.isMenuOnline) {
```
simon-the-shark, line 79: see here, we don't show the animation if the menu is online anyway, so the code from above can be deleted:
```diff
- if (sksMenuData.isMenuOnline && isLastMenuButtonClicked.value) {
-        isLastMenuButtonClicked.value = false;
-      }
```

simon-the-shark: unless I'm missing something

lib/features/sks-menu/presentation/sks_menu_screen.dart
```diff
@@ -91,45 +109,83 @@ class _SksMenuView extends StatelessWidget {
...
+                    textAlign: TextAlign.center,
+                  ),
+                ),
+              if (onShowLastMenuTap != null && isAnimationCompleted.value)
+                Padding(
+                  padding: const EdgeInsets.only(top: 16),
+                  child: ElevatedButton(
+                    onPressed: () {
+                      onShowLastMenuTap?.call();
+                    },
```
simon-the-shark, line 161: [nit]: shorter version for this:
```dart
onPressed: onShowLastMenuTap,
```

simon-the-shark (commented):
Hi! Nice work here. Personally I would probably create a new view and push it to the stack, but your approach is pretty clean too. Nothing wrong here. I actually like it.

The code looks good and pretty clean now I think. Good job there. I have only one note: we shouldn't change state during build method itself && in this case, this state change can be probably skipped (imo doesn't change anything). Apart from this, one nitpicking note and no more issues from my perspective

lib/features/sks-menu/presentation/sks_menu_screen.dart
mmzek: i think your right that it's unnecessary, I deleted it

simon-the-shark (dismissed):
LGTM

PR #462, opened 2024-12-01
Chore/455 change parking tap detection

Rejfi:
FIx problem with onTap activate after press the space above chart


tomasz-trela (approved):
LGTM!

simon-the-shark (approved):

PR #482, opened 2024-12-10
Feat/ Adjust SKS error handling && add REST API caching

simon-the-shark:
1. Adjusted error handling
  - previously we had shown sks animation with any error
  - now we show:
      - sks animation when the menu is not available (that's not really an exception, just a normal situation)
      - offline animation when we don't have internet connection
      - our `MyErrorWidget` on any other error

image
image

2. Added caching
 - we cache the data for 1 day, only when the available menu is up to date
 - we always invalidate the cache when the cached data is from the previous day (after midnight)

3. pull to refresh
    - we invalidate the cache and redo the request

Zrzut ekranu 2024-12-10 o 13 31 06

### DIFF
--- a/lib/features/sks-menu/data/repository/sks_menu_repository.dart
+++ b/lib/features/sks-menu/data/repository/sks_menu_repository.dart
@@ -3,6 +3,7 @@ import "package:riverpod_annotation/riverpod_annotation.dart";
 
 import "../../../../api_base_rest/cache/cache.dart";
 import "../../../../config/env.dart";
+import "../../../../utils/context_extensions.dart";
 import "../../../../utils/datetime_utils.dart";
 import "../models/dish_category_enum.dart";
 import "../models/sks_menu_response.dart";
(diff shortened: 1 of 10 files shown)


tomasz-trela (commented):
Good job. I wrote down one suggestion.

lib/features/sks-menu/data/repository/sks_menu_repository.dart
```diff
@@ -1,36 +1,51 @@
...
+    );
+
+    final trueMeals = sksMenuResponse.meals
+        .where((e) => e.category != DishCategory.technicalInfo)
+        .toList();
+
+    final technicalInfos = sksMenuResponse.meals
+        .where((e) => e.category == DishCategory.technicalInfo)
+        .map((e) => e.name)
+        .toList();
```
tomasz-trela, line 42: Consider changing to IList.

tomasz-trela (commented):
^

lib/features/sks-menu/data/repository/sks_menu_repository.dart
simon-the-shark: if only freezed supports this then I'm for it

simon-the-shark: done, thanks

Issue #449, opened 2024-11-30
feat/ add fullscreen mode for photos in detail views


odominiak:
also add zoom if simple enough
![Image](https://github.com/user-attachments/assets/d23624c2-2287-4b44-a7da-483f7ee66435)

PR #488, opened 2024-12-13
#449 feat/ add fullscreen mode for photos in detail views

jakubbaldyga:
Okay so, generally this works pretty nice. Although I have a problem that when the zoomable image will be shown in fullscreen, in some cases the Scaffold will only contain the image leaving black areas around it. I tried many approaches but i can't get it to work...

Correct scafold coverage:
image

Incorrect scaffold coverage (there are black areas above and under the image):
image
Notice that here there is an image with transparency, there is a greyLight color underneath it.


lib/config/ui_config.dart
```diff
@@ -89,6 +89,7 @@ abstract class WideTileCardConfig {
 }
 
 abstract class DetailViewsConfig {
+  static const heightImage = 254.0;
   static const double spacerHeight = 16;
```
mikolaj-jalocha, line 93: please specify the type of `heightImage`, imo better to rename it to: `imageHeight` (just to follow convention)

lib/features/digital_guide_view/general_info/presentation/digital_guide_view.dart
```diff
@@ -53,9 +53,12 @@ class _DigitalGuideView extends ConsumerWidget {
...
     final widgets1 = [
-      const SizedBox(height: DigitalGuideConfig.heightSmall),
-      MyCachedImage(
-        digitalGuideResponseExtended.imageUrl,
+      const SizedBox(height: DetailViewsConfig.heightImage),
+      SizedBox(
+        height: DetailViewsConfig.heightImage,
+        child: ZoomableOptimizedDirectusImage(
+          digitalGuideResponseExtended.imageUrl,
+        ),
```
mikolaj-jalocha, line 61: Is `SizedBox` declared twice needed here?

lib/widgets/detail_views/sliver_logo.dart
```diff
@@ -21,42 +21,41 @@ class SliverLogo extends StatelessWidget {
...
   final BoxFit boxfit;
   final bool noShimmeringLoading;
+
   @override
   Widget build(BuildContext context) {
     return Align(
       alignment: Alignment.bottomCenter,
       child: SizedBox.square(
-        child: ListView(
+        child: SingleChildScrollView(
```
mikolaj-jalocha, line 30: why `SingleChildScrollView` here?

lib/widgets/optimized_directus_image.dart
```diff
@@ -42,6 +31,25 @@ class OptimizedDirectusImage extends MyCachedImage {
...
+    );
+  }
+
+  Size _getImagesSize(BuildContext context, BoxConstraints constraints) {
+    final pixelsDensity = MediaQuery.devicePixelRatioOf(context);
+    return Size(
+      constraints.maxWidth * pixelsDensity,
+      constraints.maxHeight * pixelsDensity,
+    );
+  }
```
mikolaj-jalocha, line 52: imo an overkill here. I'd live it as I was before

mikolaj-jalocha (changes requested):
Really god work here! Nice and clean. I left few suggestions and questions of mine. 

Regarding problem you described in PR description, maybe it has sth to do with image size? 

@simon-the-shark if you have free time please look at it as well. Especially on part where navigator comes to play.

lib/widgets/zoomable_optimized_directus_image.dart
```diff
@@ -0,0 +1,54 @@
...
+        boxFit: boxFit,
+        noShimmeringLoading: noShimmeringLoading,
+      ),
+    );
+  }
+
+  Future<void> _showFullScreenImage(BuildContext context) async {
+    await Navigator.of(context).push(
+      MaterialPageRoute(
+        builder: (newContext) => Scaffold(
```
simon-the-shark, line 34: using `newContext` instead of just `context` can lead to mistakes

simon-the-shark (commented):
Could we make the background semi-transparent (black with some opacity) to stil see what's underneath a bit??

lib/widgets/zoomable_optimized_directus_image.dart
simon-the-shark: the convention is just to use `context` always

lib/features/digital_guide_view/general_info/presentation/digital_guide_view.dart
jakubbaldyga: Sorry there is a mistake in DetailViewsConfig.heightImage - I changed it, but it is needed as a visual gap

lib/widgets/detail_views/sliver_logo.dart
jakubbaldyga: This was changed, because in scientific club view ListView has tottally "hidden" the image behind it not allowing you to click on it. It turned out that SingleChildScrollView takes less space and thus allowng you to click on that image behind.

lib/config/ui_config.dart
jakubbaldyga: Changed -  i did it as it was done for heightBig heightSmall etc

lib/widgets/zoomable_optimized_directus_image.dart
jakubbaldyga: Changed

jakubbaldyga:
> Could we make the background semi-transparent (black with some opacity) to stil see what's underneath a bit??

Nice idea it looks great now! 
image
Unfortunately the issue with these black areas persists...
And I wonder if the transparent image had a white background it would be better:
image

mikolaj-jalocha:
> > Could we make the background semi-transparent (black with some opacity) to stil see what's underneath a bit??
> 
> Nice idea it looks great now! image Unfortunately the issue with these black areas persists... And I wonder if the transparent image had a white background it would be better: image

yeah, with transparent image white background is welcome I guess. In terms of black area, did you check the image dimensions?

simon-the-shark:
> > > Could we make the background semi-transparent (black with some opacity) to stil see what's underneath a bit??
> > 
> > 
> > Nice idea it looks great now! image Unfortunately the issue with these black areas persists... And I wonder if the transparent image had a white background it would be better: image
> 
> yeah, with transparent image white background is welcome I guess. In terms of black area, did you check the image dimensions?

yeah, so ideally we want white background but only behind the image and semi transparent on the rest of the screen. Not sure if we can achieve this here, but why not?

simon-the-shark:
What's the status here???

jakubbaldyga:
> What's the status here???

Sorry I was really busy but I got it to work now. Appareantly the issue with those black areas lied in the way the image was fetched.
https://admin.topwr.solvro.pl/assets/5fc6e11b-ad99-484b-85ff-43136b93e7a7.png?width=1080&height=2273&fit=contain
Notice that in the image returned there are black areas above and under the image.
So I changed Boxfit from contain to scaledown and it changes in the url fit=contain to fit=inside which is what I wanted from the beggining.

There was another issue with that white background since I don't see any good way to do this at all.
Generally when the image is displayed it takes all the available space. So setting background caused setting the whole screen color to white. Also it's not needed in every possible case.
To address this issue I added additional param with the default value set to disable. When enabled (for now only in logos), creates the white square background.

Here are the results now:
image
image

lib/widgets/zoomable_optimized_directus_image.dart
```diff
@@ -0,0 +1,87 @@
...
+
+  Padding _showImage() => Padding(
+        padding: const EdgeInsets.all(10),
+        child: OptimizedDirectusImage(
+          imageUrl,
+          boxFit: BoxFit.scaleDown,
+          noShimmeringLoading: true,
+        ),
+      );
+}
```
mikolaj-jalocha, line 87: So called builder methods. In our project we consider them as anti pattern.  Please see [this](https://www.reddit.com/r/FlutterDev/comments/avhvco/extracting_widgets_to_a_function_is_not_an/)

mikolaj-jalocha (changes requested):
One comment

lib/widgets/zoomable_optimized_directus_image.dart
```diff
@@ -0,0 +1,87 @@
...
+            onTap: () {
+              Navigator.of(context).pop();
+            },
+            child: InteractiveViewer(
+              minScale: 1,
+              maxScale: 3,
+              child: shouldHaveRectBackground
+                  ? _showImageWithBackground(
+                      MediaQuery.of(context).size.shortestSide,
+                    )
```
simon-the-shark, line 49: it should be `MediaQuery.sizeOf(context).shortestSide` to reduce needles rebuilds

lib/widgets/zoomable_optimized_directus_image.dart
simon-the-shark: +1

simon-the-shark:
What's the status here :)??

lib/widgets/zoomable_optimized_directus_image.dart
jakubbaldyga: Thanks. I removed those methods and inserted them to their parent method. Now there shouldn't be additional and unnecessary rebuilds.

PR #489, opened 2024-12-14
feat: evacuation screen and small refactor of file structure in digit…

tomasz-trela:
I refactored file structure a little bit. I decided to do this in tab (not in another screen) because when it's look more consistent. I know it's not the best practice in ux, but I think not many people will use this tab (so ui will benefit from this).
Simulator Screenshot - iPhone 16 Plus - 2024-12-14 at 12 56 20
I'm open for any suggestions.


lib/features/digital_guide_view/presentation/widgets/digital_guide_go_to_button.dart
```diff
@@ -0,0 +1,46 @@
+import "package:flutter/material.dart";
+
+import "../../../../config/ui_config.dart";
+import "../../../../theme/app_theme.dart";
+import "../../tabs/entraces/data/models/digital_guide_entrace.dart";
+
+class DigitalGuideGoToButton extends StatelessWidget {
```
simon-the-shark, line 7: maybe `DigitalGuideNavLink`?

lib/features/digital_guide_view/tabs/models/digital_guide_entrace.dart
simon-the-shark, line 1: THIS FILE IS DUPLICATED?!?

lib/features/digital_guide_view/presentation/widgets/digital_guide_go_to_button.dart
```diff
@@ -0,0 +1,46 @@
...
+  const DigitalGuideGoToButton({
+    super.key,
+    required this.onTap,
+    required this.entrace,
+  });
+
+  final VoidCallback onTap;
+  final DigitalGuideEntrace entrace;
```
simon-the-shark, line 15: Also I suppose this will work for other models too in the future, not only entrances? So maybe some more general parameters could be applied here? And then some `EntranceNavLink` that uses the `DigitalGuideGoToButton` /`DigitalGuideNavLink` inside. If you know what I mean

lib/features/digital_guide_view/tabs/entraces/data/models/digital_guide_entrace.dart
simon-the-shark, line 1: there is a copy of this file in tabs/models

simon-the-shark (commented):
Good job Tomuś!

I like the file structure. Only thing is that you have some duplicate trash files in my opinion, `DigitalGuideEntrace` models are twice in the filesystem and the `LocalizationExpansionTileContent` inside `localization` in not in the `tabs`. Rest look good.

Few notes left, one is a big missing one from the previous @24bartixx big PR - idk how I missed that.

I also propose to create a separate evacutation repo, but not sure if I got everything right (didn't study the responses in very much detail)


Tell me what you think!

tomasz-trela:
I can't create separate repo for evacuation because evacuation is a property of building( buildings endpoint).

simon-the-shark:
> I can't create separate repo for evacuation because evacuation is a property of building( buildings endpoint).

Your evacuation repo can use the building data provider:
```dart
@riverpod
Future<EvacuationModel> evacuationRepo(Ref ref, int id) async {
  final building = await ref.watch(getDigitalGuideDataProvider(id).future);
  final evacuationMapUrl =
      await ref.watch(getImageUrlProvider(building.evacuationMapId).future);

  // ...
  //...
  //...
  
  return EvacuationModel(
    //...
  );
}
```


or am i wrong?

tomasz-trela:
> > I can't create separate repo for evacuation because evacuation is a property of building( buildings endpoint).
> 
> Your evacuation repo can use the building data provider:
> 
> ```dart
> @riverpod
> Future<EvacuationModel> evacuationRepo(Ref ref, int id) async {
>   final building = await ref.watch(getDigitalGuideDataProvider(id).future);
>   final evacuationMapUrl =
>       await ref.watch(getImageUrlProvider(building.evacuationMapId).future);
> 
>   // ...
>   //...
>   //...
>   
>   return EvacuationModel(
>     //...
>   );
> }
> ```
> 
> or am i wrong?

Now I understand

tomasz-trela:
Done

lib/config/ui_config.dart
```diff
@@ -213,9 +213,13 @@ abstract class DigitalGuideConfig {
...
   static const borderRadiusMedium = 8.0;
   static const heightSmall = 8.0;
+  static const heightMedium = 16.0;
   static const heightBig = 24.0;
   static const heightHuge = 48.0;
+  static const mediumButtonPadding =
+      EdgeInsets.symmetric(vertical: 8, horizontal: 14);
   static const paddingMedium = 16.0;
+  static const borderColor = Color(0x8077859D);
 }
```
mikolaj-jalocha, line 223: What about defining the color in `ColorConsts`?

tomasz-trela: I can do it but I think it doesn't really matter, because we are using this color only once.

tomasz-trela: So DigitalGuideConfig is not bad place for it because scope of it is much smaller than color consts.

simon-the-shark: imo it can stay for now

simon-the-shark (approved):
basically LGTM, still I would maybe at least test the behaviour in case where we would only download this things when we open the specific tab and not all at first load, cause with time this is gonna be a lot of requests. But this require some repository rearranging - and maybe I'll handle this when I have a moment (someone else also can :P)

Merging this for now.

PR #517, opened 2025-01-06
Feat/strategic badge

simon-the-shark:
we can always change the icon, but this works for now imo. Already connected to directus and added in all places I believe.

Screenshot_1736205352
Screenshot_1736205364
Screenshot_1736205377
Screenshot_1736204899
Screenshot_1736205426


lib/features/science_clubs_view/widgets/strategic_badge.dart
```diff
@@ -0,0 +1,45 @@
...
+            child: Icon(
+              Icons.shield,
+              size: 14,
+              color: context.colorTheme.blueAzure,
+            ),
+          ),
+          Align(
+            alignment: const Alignment(0, -1 / 4),
+            child: Text(
+              "KS",
```
mikolaj-jalocha, line 37: WDYT about making it available for l10n?

mikolaj-jalocha (commented):
Nice and clean, I like the effect. Left one comment.

WDYT about prioritazing strategic circles in our app? (Making them appear 1st in every list we have)?

lib/features/science_clubs_view/widgets/strategic_badge.dart
simon-the-shark: Sure

PR #518, opened 2025-01-07
Feat/ add multiversion team section

simon-the-shark:
Screenshot_1736248918

## Limits
so it has limits of version when it stops looking nice, but at the moment not a thing to worry about

below is visualization for different version counts on a very small device
Screenshot_1736248708
Screenshot_1736248144
Screenshot_1736248407
Screenshot_1736248449


simon-the-shark:
actually I've reversed the order to more chronological
Screenshot_1736257223

PR #520, opened 2025-01-09
fix(digital-guide): ui changes to amnesties and surroundings tiles

pbonar:
### Changes:

- Small paddings changes in amnesties expansion tile
- Improving visual aspect of Surroundings expension tile
- Small changes to other parts of digital guide UI

Screenshot 2025-01-09 at 20 20 41
Screenshot 2025-01-09 at 20 2
<img width=
Screenshot 2025-01-09 at 20 21 23


simon-the-shark (commented):
All looks good to me here! Only issue is that the formatter check is failing, so u gotta run autoformatter https://github.com/Solvro/mobile-topwr/?tab=readme-ov-file#before-you-push-a-commit
```bash
dart format .
```

If u have futher issue with formatter or linter, hit me up

PR #525, opened 2025-01-11
refactor: reorganize dependencies

tomasz-trela:
I reorganized pubspec.yaml. I divided this file into categories to have cleaner look for our dependencies.

### DIFF
--- a/lib/features/bottom_nav_bar/bottom_nav_bar.dart
+++ b/lib/features/bottom_nav_bar/bottom_nav_bar.dart
@@ -20,9 +20,7 @@ class BottomNavBar extends ConsumerWidget {
         decoration: BoxDecoration(
           boxShadow: <BoxShadow>[
             BoxShadow(
-              color: Colors.black.withOpacity(
-                .08,
-              ),
+              color: Colors.black.withValues(alpha: .08),
               blurRadius: 20,
               offset: const Offset(0, -1),
             ),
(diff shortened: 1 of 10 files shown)


mikolaj-jalocha (approved):
We needed it! 
Well done!

mikolaj-jalocha:
- Everything below was tested on an iPhone 
- [Colors](https://stackoverflow.com/questions/79278159/warnings-after-upgrading-to-flutter-3-27-0-value-and-green-are-deprecated)
- [WithValues](https://stackoverflow.com/questions/79284946/why-is-withopacity-deprecated-in-flutter-3-27-0-and-what-is-its-recommended-rep)
- I didn't search for sufficient docs about onPopWithResult. it's def worth checking if on Android everything works fine (at the moment i can't do it)

simon-the-shark (commented):
basically lgtm - only question I have is if you had checked whether color values are still in the same range as before (cause some people use 0-1 floats and some 0-255 ints)

mikolaj-jalocha:
> basically lgtm - only question I have is if you had checked whether color values are still in the same range as before (cause some people use 0-1 floats and some 0-255 ints)

I've tested it empirically (and everything looks good). According to docs colors re described as values from 0 - 255 doubles.

Zrzut ekranu 2025-01-12 o 16 32 37

simon-the-shark (approved):
in god we trust

PR #526, opened 2025-01-12
Round images in adapted toilets

24bartixx:
#  Round images in adapted toilets

image

# zoomable = false in digital guide view - prevents this

image
image

# Additional info
I tried to use DigitalGuidePhotoRow in adapted toilets and I even fixed row stretching in this widget but it looked awkward because of the images ratio.

### DIFF
--- a/lib/features/digital_guide_view/presentation/digital_guide_view.dart
+++ b/lib/features/digital_guide_view/presentation/digital_guide_view.dart
@@ -64,7 +64,8 @@ class _DigitalGuideView extends ConsumerWidget {
       const SizedBox(height: DigitalGuideConfig.heightSmall),
       SizedBox(
         height: DetailViewsConfig.imageHeight,
-        child: DigitalGuideImage(id: digitalGuideData.images[0]),
+        child:
+            DigitalGuideImage(id: digitalGuideData.images[0], zoomable: false),
       ),
       HeadlinesSection(
         name: digitalGuideData.translations.plTranslation.name,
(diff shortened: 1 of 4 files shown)


lib/features/digital_guide_view/presentation/digital_guide_view.dart
```diff
@@ -64,7 +64,8 @@ class _DigitalGuideView extends ConsumerWidget {
       const SizedBox(height: DigitalGuideConfig.heightSmall),
       SizedBox(
         height: DetailViewsConfig.imageHeight,
-        child: DigitalGuideImage(id: digitalGuideData.images[0]),
+        child:
```
simon-the-shark, line 67: why is the reason for zoomable = false?

simon-the-shark (commented):
so there is not such thing as too small PR to be honest, I do make things like this in my work too.

Idk if u think some screenshot could represent what u've changed? Also could write a reason why we need zoomable = false here?

lib/features/digital_guide_view/presentation/digital_guide_view.dart
24bartixx: I provided screenshots in the PR description. Imo it looks unnaturally for this photo to be displayed in such a way

simon-the-shark:
@24bartixx tbh I don't know why the zoomable one is bad. Something to discuss in a wider group on a meeting.

When it comes to photo row - sure, I have similar conclusions, it can stay as it is for now in this view, imo it looks pretty ok

simon-the-shark (approved):

PR #533, opened 2025-01-16
Feat/connect digital guide

simon-the-shark:
1. connect digital guide to our building tiles with "i" icon
    - I allow to set manually id on the backend and the option if this is normal building, "other place" (different endpoint) or web url, cause why can't we set deeplink to sks menu or just url somewhere here too
2. I use OUR photos in the digital guide detail too, cause previously this could be confusing that the photo was different on the list and in the detail

I only added few for testing for now, @odominiak will insert rest soon

Zrzut ekranu 2025-01-17 o 00 24 28
Zrzut ekranu 2025-01-17 o 00 29 54
Zrzut ekranu 2025-01-17 o 00 30 09


mikolaj-jalocha (approved):
LGTM!

it'd be nice to add some sort of restaurant symbol to SKS's tile 😋 or deep link to parking if it's close to given building.

PR #534, opened 2025-01-17
feat(digital-guide): add support for levels

mikolaj-jalocha:

### DIFF
--- a/lib/features/digital_guide_view/tabs/rooms/data/repository/rooms_repository.dart
+++ b/lib/features/digital_guide_view/tabs/rooms/data/repository/rooms_repository.dart
@@ -16,7 +16,7 @@ Future<IList<DigitalGuideRoom>> roomsRepository(
   LevelWithRegions level,
 ) async {
   Future<DigitalGuideRoom> getDigitalGuideRoom(int roomId) async {
-    final String endpoint = "rooms/$roomId";
+    final endpoint = "rooms/$roomId";
     return ref.getAndCacheDataFromDigitalGuide(
       endpoint,
       DigitalGuideRoom.fromJson,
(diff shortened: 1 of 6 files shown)


lib/features/digital_guide_view/tabs/rooms/data/repository/rooms_repository.dart
```diff
@@ -3,24 +3,28 @@ import "package:flutter_riverpod/flutter_riverpod.dart";
...
 
 @riverpod
 Future<IList<DigitalGuideRoom>> roomsRepository(
   Ref ref,
-  DigitalGuideResponse building,
+  LevelWithRegions level,
 ) async {
-  final String endpoint = "rooms/?building=${building.id}";
+  Future<DigitalGuideRoom> getDigitalGuideRoom(int roomId) async {
+    final String endpoint = "rooms/$roomId";
```
simon-the-shark, line 19: [nit] do as you like, but letting you know that this `String` is not needed, as the types are inferred in such case

```suggestion
    final endpoint = "rooms/$roomId";
```

simon-the-shark (dismissed):
looks good to me :)). Not sure about the padding between the levels section though, at least on the ss it looks like a lot of empty space

lib/features/digital_guide_view/tabs/rooms/data/repository/rooms_repository.dart
mikolaj-jalocha: Thank you! I will apply changes, padding are to big rn, I agree

mikolaj-jalocha:
Quick fix. Unfortunately I can't test it right now with the same id as in PR's screen but should look good now. In case it is wrong, I will fix it all together in single commit with others small issues.

Issue #527, opened 2025-01-12
/feat (digital-guide) add gallery


mikolaj-jalocha:
- in the fullscreen mode carousel should be used to see 3 photos
- after click on "Zobacz wszystkie 10 zdjęć", again, some kind of carousel  or other (in your opinion suitable) option of gallery should be implemented

![Image](https://github.com/user-attachments/assets/fe4d67e9-4f90-4824-a6a3-400e7db565d2)

PR #535, opened 2025-01-17
feat(digital-guide): add gallery

tomasz-trela:
#527 
[Simulator Screen Recording - iPhone 16 Plus - 2025-01-17 at 19.42.00.mp4.zip](https://github.com/user-attachments/files/18458572/Simulator.Screen.Recording.-.iPhone.16.Plus.-.2025-01-17.at.19.42.00.mp4.zip)
Zrzut ekranu 2025-01-17 o 19 50 21

Sorry for sending video in zip but I don't have time and file is too big.

### DIFF
--- a/lib/features/digital_guide_view/presentation/widgets/digital_guide_carousel.dart
+++ b/lib/features/digital_guide_view/presentation/widgets/digital_guide_carousel.dart
@@ -1,0 +1,77 @@
+import "dart:async";
+
+import "package:carousel_slider_plus/carousel_slider_plus.dart";
+import "package:fast_immutable_collections/fast_immutable_collections.dart";
+import "package:flutter/material.dart";
+
+import "../../../../config/ui_config.dart";
+import "digital_guide_image.dart";
+
+class DigitalGuideCarouselWithIndicator extends StatefulWidget {
+  const DigitalGuideCarouselWithIndicator({super.key, required this.imgListId});
+
+  final IList<int> imgListId;
+
+  @override
+  State<StatefulWidget> createState() {
+    return _CarouselWithIndicatorState();
+  }
+}
+
+class _CarouselWithIndicatorState
+    extends State<DigitalGuideCarouselWithIndicator> {
+  int _current = 0;
+  final CarouselSliderController _controller = CarouselSliderController();
+
+  @override
+  Widget build(BuildContext context) {
+    return Dialog(
+      insetPadding: EdgeInsets.zero,
+      backgroundColor: Colors.transparent,
+      child: Wrap(
+        children: [
+          CarouselSlider(
+            items: widget.imgListId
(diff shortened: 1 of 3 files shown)


lib/features/digital_guide_view/presentation/widgets/digital_guide_carousel.dart
```diff
@@ -0,0 +1,78 @@
...
+
+  @override
+  State<StatefulWidget> createState() {
+    return _CarouselWithIndicatorState();
+  }
+}
+
+class _CarouselWithIndicatorState
+    extends State<DigitalGuideCarouselWithIndicator> {
+  int _current = 0;
```
simon-the-shark, line 23: so it's a matter of convebtion, but for some time we prefer using flutter hooks over statefullwidget.  so here u could just use `useState` from there

lib/features/digital_guide_view/presentation/widgets/digital_guide_carousel.dart
```diff
@@ -0,0 +1,78 @@
...
+  final CarouselSliderController _controller = CarouselSliderController();
```
simon-the-shark, line 24: one note for the future, I've checked it and that's not the case here, but:

So if you have anything `XController` in Flutter, this often means that u have to be careful about memory leaks. Cause usually `AnythingController` in flutter should be disposed with `controller.dispose()` in the widget's dispose method (or using some hooks) and it's often ommited, even by the flutter team itself inside of the framework :P

But here we don't have such option. So either this is not needed, or the package's author is causing memory leaks xdd

simon-the-shark (commented):
so this looks good Tomuś :)

1. First question is, did you considered using new native `CarouselView` from the flutter framework: https://api.flutter.dev/flutter/material/CarouselView-class.html :P - maybe this is better, idk, just leaving this as a note for the future

2. I would opt for a flutter hooks instead of stateful widget here

simon-the-shark:
About the UX, could we also open this dialog when clicked on one of the 3 visible photos too?
Zrzut ekranu 2025-01-17 o 21 50 56

tomasz-trela:
> so this looks good Tomuś :)
> 
> 1. First question is, did you considered using new native `CarouselView` from the flutter framework: https://api.flutter.dev/flutter/material/CarouselView-class.html :P - maybe this one is better, idk, just leaving this as a note for the future
> 2. I would opt for a flutter hooks instead of stateful widget here

I considered this and I think it's better to use this package because there is more built in functionalities, like this effect which enlarges center-panel and auto-play effect. It's not that difficult to do it myself, but now I don't have much time so I decided to this option. I think the biggest con of this solution is the fact that this widget can cause memory leaks.

simon-the-shark (approved):
lgtm :))

PR #537, opened 2025-01-18
feat(digital-guide): add accessibility modes dialog

simon-the-shark:
Nagranie z ekranu Jan 18 2025 (1)

so when it comes to code, i went kinda crazy...

cause this could have been easily hardcoded, and I'm not sure if the layers of abstractions are not too much a bit for the use case. Also some functional programming concepts were used, allowing us to define any number of custom accessibility modes and nest them into each other in any way we want. So please tell me if this is still readable or would you prefer something more hardcoded in a way, and not fully abstract.

this could be usefull to read before the review: https://dart.dev/language/class-modifiers#sealed

### DIFF
--- a/lib/features/digital_guide_view/tabs/accessibility_dialog/data/modes.dart
+++ b/lib/features/digital_guide_view/tabs/accessibility_dialog/data/modes.dart
@@ -1,0 +1,48 @@
+// https://dart.dev/language/class-modifiers#sealed
+sealed class AccessibilityMode {
+  const AccessibilityMode();
+}
+
+// this mode's state depends on the state of its children
+sealed class ModeWithChildren extends AccessibilityMode {
+  const ModeWithChildren(this.children);
+  final List<AccessibilityMode> children;
+}
+
+// this mode's state is stored locally in shared preferences
+sealed class ModeWithKey extends AccessibilityMode {
+  const ModeWithKey(this.sharedPrefsKey);
+  final String sharedPrefsKey;
+}
+
+class MotorImpairment extends ModeWithKey {
+  const MotorImpairment() : super("_prefs_accessibility_motor_impairment");
+}
+
+class Blind extends ModeWithKey {
+  const Blind() : super("_prefs_accessibility_blind");
+}
+
+class LowVision extends ModeWithKey {
+  const LowVision() : super("_prefs_accessibility_low_vision");
+}
+
+class SensorySensitivity extends ModeWithKey {
+  const SensorySensitivity()
+      : super("_prefs_accessibility_sensory_sensitivity");
+}
+
+class CognitiveImpairment extends ModeWithKey {
+  const CognitiveImpairment()
+      : super("_prefs_accessibility_cognitive_impairment");
+}
+
+class VisualImpairment extends ModeWithChildren {
+  const VisualImpairment()
+      : super(
+          const [
+            LowVision(),
+            Blind(),
+          ],
+        );
+}
--- a/lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/checkboxes_list.dart
+++ b/lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/checkboxes_list.dart
@@ -1,0 +1,35 @@
+import "package:flutter/material.dart";
+import "package:flutter_hooks/flutter_hooks.dart";
+
+import "../business/top_level_modes.dart";
+import "../data/modes.dart";
+import "mode_checkbox.dart";
+
+class _SubModePadding extends Padding {
+  const _SubModePadding({super.child})
+      : super(
+          padding: const EdgeInsets.only(left: 25),
+        );
+}
+
+class CheckboxesList extends HookWidget {
+  const CheckboxesList({super.key});
+
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      children: [
+        for (final mode in topLevelModes)
+          switch (mode) {
+            ModeWithChildren() => Column(
+                children: [
+                  for (final subMode in mode.children)
+                    _SubModePadding(child: ModeCheckbox(subMode)),
+                ],
+              ),
+            ModeWithKey() => ModeCheckbox(mode),
+          },
+      ],
+    );
+  }
+}
--- a/lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/mode_checkbox.dart
+++ b/lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/mode_checkbox.dart
@@ -1,0 +1,40 @@
+import "dart:async";
+
+import "package:flutter/material.dart";
+import "package:flutter_riverpod/flutter_riverpod.dart";
+
+import "../../../../../theme/app_theme.dart";
+import "../business/accessibility_mode_service.dart";
+import "../data/modes.dart";
+import "labels.dart";
+
+class ModeCheckbox extends ConsumerWidget {
+  const ModeCheckbox(this.mode, {super.key});
+  final AccessibilityMode mode;
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final value = ref.watch(accessibilityModeServiceProvider(mode));
+
+    // ignore: avoid_positional_boolean_parameters
+    void onChanged(bool? value) {
+      if (value != null) {
+        unawaited(
+          ref
+              .read(accessibilityModeServiceProvider(mode).notifier)
+              .setMode(newValue: value),
+        );
+      }
+    }
+
+    return CheckboxListTile(
+      dense: true,
+      controlAffinity: ListTileControlAffinity.leading,
+      title: Text(
+        mode.localizedLabel(context),
+        style: context.aboutUsTheme.body,
+      ),
+      value: value.value ?? false,
+      onChanged: onChanged,
+    );
+  }
+}
--- a/lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
+++ b/lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
@@ -1,0 +1,151 @@
+import "package:flutter/material.dart";
+
+import "../../../../../config/ui_config.dart";
+import "../../../../../theme/app_theme.dart";
+import "../../../../../utils/context_extensions.dart";
+
+// pure UI, no logic, just a nice dialog with a title, subtitle and a child
+class RedDialog extends StatelessWidget {
+  final String title;
+  final String subtitle;
+  final Widget child;
+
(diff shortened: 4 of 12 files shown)


lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
```diff
@@ -0,0 +1,114 @@
...
+              Padding(
+                padding: const EdgeInsets.all(20).copyWith(bottom: 0),
+                child: Row(
+                  mainAxisAlignment: MainAxisAlignment.spaceBetween,
+                  children: [
+                    Text(
+                      title,
+                      style: context.textTheme.headline.copyWith(height: 1.4),
+                    ),
+                    IconButton(
```
24bartixx, line 58: The text and icon in the row might overlap truncating the text if the necessary width is lower than the available width. It might be a good idea to make it bulletproof.

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
```diff
@@ -0,0 +1,114 @@
...
+                    ),
+                    shape: RoundedRectangleBorder(
+                      borderRadius: FilterConfig.radius,
+                    ),
+                  ),
+                  child: Row(
+                    mainAxisAlignment: MainAxisAlignment.center,
+                    children: [
+                      Text(
+                        context.localize.apply,
```
24bartixx, line 101: Can't we simply wrap the Text inside the Center Widget instead of the Row Widget? I might be wrong but I believe the Row purpose is different than to center sth out.

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/mode_checkbox.dart
```diff
@@ -0,0 +1,39 @@
...
+              .setMode(newValue: value),
+        );
+      }
+    }
+
+    return CheckboxListTile(
+      controlAffinity: ListTileControlAffinity.leading,
+      title: Text(
+        mode.localizedLabel(context),
+        style: context.aboutUsTheme.body,
```
24bartixx, line 33: (NIT) The size of the checkbox text bothers me optically a little as it is slightly bigger than the light grey text above checboxes. However, it might be only my preference idk

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/checkboxes_list.dart
```diff
@@ -0,0 +1,30 @@
...
+class _SubModePadding extends Padding {
+  const _SubModePadding({super.child})
+      : super(
+          padding: const EdgeInsets.only(left: 25),
+        );
+}
+
+class CheckboxesList extends HookWidget {
+  const CheckboxesList({super.key});
+
```
24bartixx, line 16: (NIT) Possibly I am wrong but I think that the space between checkboxes is slightly too big. Maybe try to decrease this size specifying a parameter of CheckboxListTile or a parameter of the appropriate column.

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/checkboxes_list.dart
```diff
@@ -0,0 +1,30 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return const Column(
+      children: [
+        ModeCheckbox(MotorImpairment()),
+        ModeCheckbox(VisualImpairment()),
+        _SubModePadding(child: ModeCheckbox(Blind())),
+        _SubModePadding(child: ModeCheckbox(LowVision())),
+        ModeCheckbox(SensorySensitivity()),
```
24bartixx, line 25: Couldn't we handle sub checkboxes in ModeCheckbox based on info of passed mode? Enumerating all of the subnodes in the correct order using a list seems like hardcoding a little bit. What I mean is to somehow generate sub checkboxes in ModeCheckbox so that they are for sure divided. Did I explain it correctly? If so WDYT?

24bartixx (changes requested):
The concept of abstract layers seems okay to me. We could have done it like this or simply hardcode and both options have their nuances. We are programmers and we like abstract things, right? haha

Besides, really good work <3

mikolaj-jalocha:
This looks like significant part of code :-) I will look into that carefully but currently short on time. Will do that till tomorrow's weekly ;)

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
simon-the-shark: yeah, you're probably right, the row is a leftover from somewhere else where we had icon too

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/checkboxes_list.dart
simon-the-shark: yeah, I've though about it too

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
simon-the-shark: good point

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/checkboxes_list.dart
simon-the-shark: good point, I've made it more dense

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/mode_checkbox.dart
simon-the-shark: I've basically followed design specs 1:1, but there's something about it

simon-the-shark: but no, this actually is the same font size

lib/features/digital_guide_view/tabs/accessibility_dialog/data/modes.dart
mikolaj-jalocha, line 1: I like this. In Android native environment this "style" of coding is extremely common making features less painful to extend in the future.

mikolaj-jalocha (approved):
LGTM! Clean, understandable code that will make possible future upgrades not painful at all (everything we'd need to do is to add next class to `sealed` wrapper). 

I'm huge fan of this although I suppose for totally fresh devs such solution is complex and not obvious to understand at first glance. In our app, that's in constant development, architecture that enable us to freely extend existing features is imo priority and there's not such a thing as an overkill.

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
```diff
@@ -0,0 +1,156 @@
...
+  @override
+  Widget build(BuildContext context) {
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        Padding(
+          padding: const EdgeInsets.symmetric(horizontal: 20),
+          child: Text(
+            subtitle,
+            style: context.aboutUsTheme.body.copyWith(
```
thesun901, line 81: Someday, someone will be really surprised that after some theme changes in aboutUs some random dialog in buildings information will also change :D 

(we should organise our themes and other UIConfigs before we end digital_guide section)

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/mode_checkbox.dart
```diff
@@ -0,0 +1,40 @@
...
+        );
+      }
+    }
+
+    return CheckboxListTile(
+      dense: true,
+      controlAffinity: ListTileControlAffinity.leading,
+      title: Text(
+        mode.localizedLabel(context),
+        style: context.aboutUsTheme.body,
```
thesun901, line 34: the same theme issues imo, but overall its acceptable for now

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
```diff
@@ -0,0 +1,156 @@
...
+          // we're saving the changes in real time anyway
+          Navigator.of(context).pop();
+        },
+        style: ElevatedButton.styleFrom(
+          backgroundColor: context.colorTheme.orangePomegranadeLighter,
+          elevation: 2,
+          padding: const EdgeInsets.symmetric(
+            vertical: 10,
+            horizontal: 20,
+          ),
```
thesun901, line 113: [NIT] kind of lots of magic numbers, especially padding is confusing and could be separated to DigitalGuideConfig

thesun901 (approved):
Implementing it in many levels of abstraction is really good sollution IMO. Everything looks clean. I left some nitpicks about the styling (looks like paddings will haunt us forever :D) But I dont think it's priority to change that now

Very impressive work!

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
thesun901: Btw I changed my mind it shouldn't be in DigitalGuideConfig, but in some new DialogsConfig - it is very probable we will use dialogs styled like this in the future so the consistency rules :D

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
simon-the-shark: also the paddings need some common design system

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/mode_checkbox.dart
simon-the-shark: yeah, it's a problem but in half of the app

simon-the-shark: if you want, you can try to unify some shit

simon-the-shark: but imo the paddings are even more crucial than fonts

lib/features/digital_guide_view/tabs/accessibility_dialog/presentation/red_dialog.dart
simon-the-shark: this padding actually is about button and it's shared with filters button, so we can drop it to already existing config or smth

Issue #541, opened 2025-01-19
/feat(digital-guide) add "szatnia" section


mikolaj-jalocha:
- Add tile for cloakroom if given building has it
- Detailed mock can be found on our Figma. Please follow the design carefully

Below I placed ss showing how it was resolved on website

![Image](https://github.com/user-attachments/assets/fa3259fa-2628-44ed-b5bf-5ccc78f12baf)

PR #547, opened 2025-01-23
Add dressing rooms to digital guide

mikolaj-jalocha:
#541 

- 1:1 changes as with lodges (I also changed working hours to be displayed only when information's available)
- I was thinking for a moment whether to create single UI that'd handle lodges and dressing rooms but finally decided just to copy and adjust existing solution. Imo digital guide structure is complicated, there's no point in complicating it even more, minor code repetition is not a sin in this scenario I believe 😆 But let me know WDYT


simon-the-shark (approved):
lgtm :))

So little repetition is not a problem. It's actually considered a good production code, cause abstraction layers are not good when they don't bring us much value. So good decision here imo.

PR #548, opened 2025-01-25
Add lifts to digital guide

mikolaj-jalocha:
#493

### DIFF
--- a/lib/features/digital_guide_view/tabs/lifts/presentation/digital_guide_lift_detail_view.dart
+++ b/lib/features/digital_guide_view/tabs/lifts/presentation/digital_guide_lift_detail_view.dart
@@ -6,12 +6,11 @@ import "package:flutter_riverpod/flutter_riverpod.dart";
 import "../../../../../config/ui_config.dart";
 import "../../../../../theme/app_theme.dart";
 import "../../../../../utils/context_extensions.dart";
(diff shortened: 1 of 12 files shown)


lib/features/digital_guide_view/tabs/lifts/presentation/digital_guide_lift_detail_view.dart
```diff
@@ -0,0 +1,84 @@
...
+      DigitalGuidePhotoRow(imagesIDs: lift.imagesIds.toIList()),
+    ];
+    return HorizontalSymmetricSafeAreaScaffold(
+      appBar: DetailViewAppBar(
+        actions: [AccessibilityButton()],
+      ),
+      body: Padding(
+        padding: const EdgeInsets.all(DigitalGuideConfig.paddingMedium),
+        child: ListView.builder(
+          physics: const NeverScrollableScrollPhysics(),
```
simon-the-shark, line 76: are we sure this isn't supoosed to be scrollable?

simon-the-shark (commented):
Good job :)). Only issue is that this photo looks weird. So here I think we can do like in the toilets view, cause both toilets and elevators will usually have vertical photos and not horizontal one. So would opt for a photo more like this:
image

mikolaj-jalocha:
Good point!

simon-the-shark (approved):

PR #557, opened 2025-01-25
Add planner advert

mikolaj-jalocha:
- whole section is clickable
- when disabled, nothing is shown 
- I didn't implement adverts in other places (I don't know exactly where and wanted to sort out the most important for now) but if you guys think it belongs to the scope of this task I will do that.
- I reused `technicalMessage` widget. Whole operation wasn't intrusive and easy to perform. I decided not to implement separate widget for that, new improvements seems useful afterall.

### DIFF
--- a/lib/features/home_view/home_view.dart
+++ b/lib/features/home_view/home_view.dart
@@ -5,6 +5,7 @@ import "package:flutter/material.dart";
 import "../../config/ui_config.dart";
 import "../../theme/app_theme.dart";
 import "../academic_calendar/widgets/academic_calendar_consumer.dart";
+import "../planner_advert/widgets/planner_advert_widget.dart";
 import "keep_alive_home_view_providers.dart";
 import "widgets/buildings_section/buildings_section.dart";
 import "widgets/logo_app_bar.dart";
@@ -23,6 +24,7 @@ class HomeView extends StatelessWidget {
         padding: EdgeInsets.only(top: 12, bottom: 4),
         child: NavActionsSection(),
       ),
+      PlannerAdvertBanner(),
       const ScienceClubsSection(),
       const BuildingsSection(),
     ].lock;
@@ -35,7 +37,7 @@ class HomeView extends StatelessWidget {
         child: ListView.separated(
           itemBuilder: (context, index) => sections[index],
           separatorBuilder: (context, index) => SizedBox(
-            height: index == 1 ? 0 : HomeViewConfig.paddingMedium,
+            height: index == 1 || index == 2 ? 0 : HomeViewConfig.paddingMedium,
           ),
           itemCount: sections.length,
         ),
--- a/lib/features/planner_advert/widgets/planner_advert_widget.dart
+++ b/lib/features/planner_advert/widgets/planner_advert_widget.dart
@@ -1,0 +1,56 @@
+import "dart:async";
+
+import "package:flutter/material.dart";
+import "package:flutter_riverpod/flutter_riverpod.dart";
+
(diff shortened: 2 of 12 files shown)


lib/features/planner_advert/widgets/planner_advert_widget.dart
```diff
@@ -0,0 +1,56 @@
...
+              title: data.title,
+              message: data.description,
+              alertType: AlertType.info,
+              icon: Icon(
+                Icons.open_in_new_rounded,
+                color: context.colorTheme.whiteSoap,
+              ),
+              onTap: () async {
+                unawaited(ref.launch(data.url));
+              },
```
tomasz-trela, line 51: You can use the arrow syntax here without extra brackets:

```dart
async () => unawaited(ref.launch(data.url))
```

lib/features/planner_advert/widgets/planner_advert_widget.dart
```diff
@@ -0,0 +1,56 @@
...
+class _PlannerAdvertBanner extends ConsumerWidget {
+  const _PlannerAdvertBanner(
+    this.data,
+  );
+
+  final PlannerAdvertContent data;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    return data.isEnabled
```
tomasz-trela, line 36: I would here reverse the condition to put smaller part of code first
```dart
!data.isEnabled
```

tomasz-trela (changes requested):
LGTM , but there is 2 small improvements to do.

lib/features/planner_advert/widgets/planner_advert_widget.dart
mikolaj-jalocha: Good point!

lib/features/planner_advert/widgets/planner_advert_widget.dart
mikolaj-jalocha: Sure, good point.

lib/features/home_view/home_view.dart
```diff
@@ -35,7 +37,7 @@ class HomeView extends StatelessWidget {
         child: ListView.separated(
           itemBuilder: (context, index) => sections[index],
           separatorBuilder: (context, index) => SizedBox(
-            height: index == 1 ? 0 : HomeViewConfig.paddingMedium,
+            height: index == 1 || index == 2 ? 0 : HomeViewConfig.paddingMedium,
```
simon-the-shark, line 40: did you check how this will behave when the advert is turned off?

simon-the-shark (commented):
so basically looks good to me, but I'd like you to check how the paddings will behave when the AD is turned off

lib/features/home_view/home_view.dart
mikolaj-jalocha: Yes I did, imo not bad (please see the pic below). If you like, I will reduce it a bit, but one question arise. Do you have an idea how to do it right? Currently I can't see other options that manually added spacers (besides making HomeScreen Consumer widget, imo wrong way).

simon-the-shark: no better ideas. If you don't want any `if`s in the separator, then u can give padding directly to children, just by wrapping them in `Padding`, but here this will look bad. Imo the current solution is ok

simon-the-shark: also the padding seems fine

Rei-x:
@mikolaj-jalocha planner -> planer 😭

Issue #559, opened 2025-01-26
chore(tests): write tests for converting colors


mikolaj-jalocha:
- write tests that covers color change from string to hex

PR #562, opened 2025-01-27
Tests for hex colors

mikolaj-jalocha:
#559 
- as a reparation for my mistake made when refactoring colors, I decided to write some tests
- test checks whether hex value returned from our custom `HexColor` class is  for each color defined in `ColorsConsts` exactly as it should be
- there were some discrepancies that test revealed. I fixed them.
- if you have some ideas about future test that could be written, please hmu. I will do that with pleasure.


tomasz-trela:
> you can also add running tests to github actions

good idea

PR #570, opened 2025-02-12
/feat(digital-guide): add "dojazd" section

mmzek:
I added a "dojazd" section and used filter in repository due to problems with API endpoint.
Obraz 1
Obraz 2
Obraz 4


mmzek:
I think it should work now

lib/features/digital_guide_view/tabs/transportation/presentation/private_transport_detail_view.dart
```diff
@@ -0,0 +1,112 @@
...
+            transportationResponse.translations.pl.areBicycleStandsComment!,
+          if (transportationResponse.translations.pl.isCityBikeStationComment !=
+              "")
+            transportationResponse.translations.pl.isCityBikeStationComment! +
+                (transportationResponse.cityBikeStationDistance != null
+                    ? context.localize.distance +
+                        transportationResponse.cityBikeStationDistance
+                            .toString() +
+                        context.localize.meters
+                    : ""),
```
tomasz-trela, line 84: I think you can create local value here like to reduce code repetition and make code much cleaner
```dart
final plTransl = transportationResponse.translations.pl;
```
this is only example name of this local value.

lib/features/digital_guide_view/tabs/transportation/presentation/public_transport_detail_view.dart
```diff
@@ -0,0 +1,147 @@
...
+                    : ""),
+          if (transportationResponse
+              .areNotPassTrafficLightsFromStopToEntryAltRoad!)
+            context.localize.no_pass_traffic_lights +
+                (transportationResponse.translations.pl
+                            .areNotPassTrafficLightsFromStopToEntryAltRoadComment !=
+                        ""
+                    ? transportationResponse.translations.pl
+                        .areNotPassTrafficLightsFromStopToEntryAltRoadComment!
+                    : ""),
```
tomasz-trela, line 121: same here and everywhere else

lib/features/digital_guide_view/tabs/transportation/presentation/transportation_expansion_tile_content.dart
```diff
@@ -0,0 +1,73 @@
...
+      child: Column(
+        crossAxisAlignment: CrossAxisAlignment.stretch,
+        children: [
+          DigitalGuideNavLink(
+            onTap: () async {
+              await ref.navigatePublicTransportDetails(transportation);
+            },
+            text: context.localize.public_transport,
+          ),
+          const SizedBox(height: 16),
```
tomasz-trela, line 62: Use value from ui-config.

tomasz-trela (changes requested):
I left a few notes. Please ping me again when you apply this requirements.

lib/features/digital_guide_view/tabs/transportation/presentation/private_transport_detail_view.dart
```diff
@@ -0,0 +1,112 @@
...
+          if (transportationResponse.isBicyclePathLeadToBuilding != null)
+            "${context.localize.bicycle_path_to_building}${transportationResponse.translations.pl.isBicyclePathLeadToBuildingComment != "" ? transportationResponse.translations.pl.isBicyclePathLeadToBuildingComment! : ""} ${transportationResponse.distanceToBicyclePath != null ? context.localize.distance + transportationResponse.distanceToBicyclePath.toString() + context.localize.meters : ""} ${transportationResponse.translations.pl.isBicyclePathLeadClearlySeparatedComment != "" ? transportationResponse.translations.pl.isBicyclePathLeadClearlySeparatedComment! : ""}",
```
mikolaj-jalocha, line 86: What's that?🧐

lib/features/navigator/app_router.dart
```diff
@@ -120,6 +123,8 @@ class AppRouter extends RootStackRouter {
           path: "/digital-guide/:id/lift-details",
           page: DigitalGuideLiftDetailRoute.page,
         ),
+        AutoRoute(page: PublicTransportDetailRoute.page),
+        AutoRoute(page: PrivateTransportDetailRoute.page),
```
mikolaj-jalocha, line 127: These pages looks almost identical. WDYT about making it as one screen? Maybe with different title

lib/l10n/app_pl.arb
```diff
@@ -309,6 +309,16 @@
   "dimensions" : "Wymiary",
   "max_capacity" : "Maksymalny udźwig",
   "generic_error_message": "Upsss...Coś poszło nie tak",
-  "you_can_adjust": "Możesz dostosować informacje pod swoje specjalne potrzeby",
-  "navigate_to_building": "Nawiguj do budynku"
+  "navigate_to_building": "Nawiguj do budynku",
+  "public_transport": "Publiczny środek komunikacji miejskiej (autobus, tramwaj)",
+  "private_transport": "Prywatny środek transportu (samochód, rower)",
+  "distance": " W odległości ",
+  "meters": " m. ",
```
mikolaj-jalocha, line 316: I don't think this is necessary XD or maybe in Cyrillic it would look different, so let it stay. I'd consider change of the name to sth like `meters_abbrev`, because at first glance I interpreted it like "metry" or "meters" not "m"

mikolaj-jalocha (changes requested):
Please consider especially my request to introduce variable for readability improvement. This will make debugging and review, and potential improvement a lot easier.

simon-the-shark (commented):
so my fellow techleads have probably mentioned everything

lib/features/digital_guide_view/tabs/transportation/presentation/transportation_detail_view.dart
```diff
@@ -0,0 +1,146 @@
...
+  }
+}
+
+extension DistanceExtension on BuildContext {
+  String distance(double? dist) {
+    return (dist != null
+        ? localize.distance + dist.toString() + localize.meters_abbrev
+        : "");
+  }
+}
```
mikolaj-jalocha, line 146: Please follow our naming convention for extensions.

lib/features/digital_guide_view/tabs/transportation/presentation/transportation_detail_view.dart
```diff
@@ -0,0 +1,146 @@
...
+                  plTransl.areBicycleStandsComment!,
+            plTransl.isCityBikeStationComment! +
+                context.distance(transportation.cityBikeStationDistance),
+            if (transportation.isBicyclePathLeadToBuilding != null)
+              context.localize.bicycle_path_leads_to_building +
+                  plTransl.isBicyclePathLeadToBuildingComment! +
+                  context.distance(transportation.distanceToBicyclePath) +
+                  plTransl.isBicyclePathLeadClearlySeparatedComment!,
+          ].lock,
+        ),
```
mikolaj-jalocha, line 114: It's good that now we have only one screen for handling 2 similar destinations. WDYT about moving those 2 lists to separate variables (and then display them) to even more improve readability?

mikolaj-jalocha (changes requested):
2 finals touches

lib/features/digital_guide_view/tabs/transportation/presentation/transportation_detail_view.dart
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  final DigitalGuideTransportation transportation;
+  final bool isPublic;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final plTransl = transportation.translations.pl;
+    final publicTransport = [
+      if (plTransl.nearestPublicTransportStop!.isNotEmpty)
```
simon-the-shark, line 28: my main issue with this code now is that there is lots of force unwrapping - which is not type safe nor nice. We should generally avoid this (`!`) when possible.

if you're sure that this can never be null, you just should change the type in the freezed model from `String?` to `String`

if you're not sure that's the case, then you should handle it here more gracefully and in more safe and typesafe way. (e.g. with a default value)

simon-the-shark (commented):
generally nicely done, but I have still one issue with this

lib/features/digital_guide_view/tabs/transportation/presentation/transportation_detail_view.dart
mmzek: youre right at first, when I tried changing those variables in the model to non-nullable ones, I was getting an error, but clearly, the issue was caused by something else because now it works fine I think

simon-the-shark (dismissed):

PR #572, opened 2025-02-13
bug: Sks menu button fix

jakubbaldyga:
Fixed that unclickable button issue. The problem was that"Pokaż ostatnio dostępne menu" button was under the image, but i managed to make it clickable and I just moved the whole section up. This is the outcome:
image
image
image

And without the button, if the latest menu is not available:
image
image
image

### DIFF
--- a/lib/features/sks_menu/presentation/sks_menu_screen.dart
+++ b/lib/features/sks_menu/presentation/sks_menu_screen.dart
@@ -142,7 +142,7 @@ class _SKSMenuUnavailableAnimation extends HookWidget {
         ) *
         0.6;
 
-    const animationTopOffset = -0.2;
+    const animationTopOffset = -0.1;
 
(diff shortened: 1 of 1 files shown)


jakubbaldyga:
hmmm, that flutter analyze doesnt appear on my side, but I think I can fix this...

jakubbaldyga:
dunno, they say here https://pub.dev/documentation/wiredash/latest/wiredash/WiredashController/show.html that this show uses future, but on my side locally it doesnt XD. There is probably some issue with pubsec?

This is my version of that function.


The fix would be probably adding async and await here (lib/features/analytics/show_feedback_tile.dart): 

onTap: () async {
        await Wiredash.of(context).show(
          options: WiredashFeedbackOptions(
            labels: context.labels,
          ),
        );
      },

PR #577, opened 2025-02-19
feat: add schody section

thesun901:
Typical digital guide work: added model, repository, presentation and translations for "Schody" section. Not all accesibility comments are implemented yet (just added them to show how it would look like)
image

Btw I think photo row should be on bottom of the screen but I didn't mess up with it for now cause im not UI/UX :D


lib/features/digital_guide/tabs/rooms/stairs/models/room_stairs_response.dart
```diff
@@ -0,0 +1,63 @@
...
+    required String isStairsSurfaceShiny,
+    required String isGoodStairsWallContrast,
+    required String areStepsEdgesMarkedInContrastedWay,
+    required String areAllStepsHaveSameDimensions,
+    required String isRailing,
+    required double railingHeight,
+    required String isGoodRailingContrast,
+    required String isRoundCrossSectionRailing,
+    required String isRailingOnRamp,
+    required String isEdgeBetweenRampRowEntranceMarked,
```
thesun901, line 24: you might wonder why I didnt convert String to Bool... well they have another state "Unknown" not just "True" or "False". On true or false comment will be shown, on unknown string connected to this property is empty

thesun901: "they" I mean digital guide API

simon-the-shark: in such case, I would cast them to some enum

lib/features/digital_guide/tabs/rooms/stairs/utils/room_stairs_comment_list.dart
```diff
@@ -0,0 +1,46 @@
+import "package:fast_immutable_collections/fast_immutable_collections.dart";
+import "package:flutter/widgets.dart";
+import "../../../../../../utils/context_extensions.dart";
+
+import "../models/room_stairs_response.dart";
+
+extension RoomStairsCommentListX on BuildContext {
```
simon-the-shark, line 7: i like this extraction with the strings here

lib/features/digital_guide/tabs/rooms/stairs/utils/room_stairs_comment_list.dart
```diff
@@ -0,0 +1,46 @@
...
+      "${localize.room_stairs_is_stairs_surface_shiny(
+        roomStairsResponse.isStairsSurfaceShiny.toLowerCase(),
+      )} ${comments.isStairsSurfaceShinyComment}",
+      "${localize.room_stairs_are_all_steps_have_same_dimensions(
+        roomStairsResponse.areAllStepsHaveSameDimensions.toLowerCase(),
+      )} ${comments.areAllStepsHaveSameDimensionsComment}",
+      "${localize.room_stairs_is_railing(
+        roomStairsResponse.isRailing.toLowerCase(),
+      )} ${comments.isRailingComment}",
+    ].where((c) => c.trim().isNotEmpty).toIList();
```
simon-the-shark, line 42: when can this strings be empty?

simon-the-shark (commented):
real good job here, just few nitpicks

lib/features/digital_guide/tabs/rooms/stairs/utils/room_stairs_comment_list.dart
thesun901: situation when for example isRailing is "unknown" and has no additional comment

lib/features/digital_guide/tabs/rooms/stairs/models/room_stairs_response.dart
thesun901: it may be an idea but those enums will be converted once again to string because .arb file accepts only some numbers and strings as agruments co we would cast string -> enum -> string

if there is reason for this (ig smaller cache size) then I can do it this way

lib/features/digital_guide/tabs/rooms/stairs/presentation/room_stairs_content.dart
```diff
@@ -0,0 +1,65 @@
...
+  const _RoomStairsContent({required this.roomStairsResponse});
+
+  final RoomStairsResponse roomStairsResponse;
+
+  @override
+  Widget build(BuildContext context, WidgetRef ref) {
+    final IList<String> roomStairsComments =
+        context.getRoomStairsCommentsList(roomStairsResponse);
+
+    final IList<String> roomStairsAccessibilityComments =
```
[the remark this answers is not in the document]
thesun901: yeah you are right it works without typing - will commit that soon

PR #582, opened 2025-02-20
Feat/update repo with accessibility

tomasz-trela:
:(


lib/features/digital_guide/data/models/digital_guide_response.dart
```diff
@@ -60,6 +60,18 @@ class DigitalGuideResponse with _$DigitalGuideResponse {
     @JsonKey(name: "location_map") required int locationMapId,
     @JsonKey(name: "levels") required List<int> levelsIndices,
     @JsonKey(name: "access") required int accessId,
+    @JsonKey(name: "accessibility_level_for_motor_disability")
```
simon-the-shark, line 63: so this should be done with `FieldRename.snake`

PR #602, opened 2025-02-25
fix(digital-guide): borders bug in new tabs

tomasz-trela:


tomasz-trela:
I fixed bug I was fixing before but in new tabs

simon-the-shark (approved):

PR #645, opened 2025-03-11
fix: convert Polish time from API to UTC in sks chart model

24bartixx:
I convert Polish time to user local time in sks chart model

image
image

----

> [!IMPORTANT]
> Convert Polish time to local time in `SksChartData` and `SksUserData` models using `fromJsonToLocalTime`.
> 
>   - **Behavior**:
>     - Convert Polish time to local time using `fromJsonToLocalTime` in `SksChartData` and `SksUserData` models.
>     - Update `externalTimestamp` field in `sks_chart_data.dart` and `sks_user_data.dart` to use `fromJsonToLocalTime`.
>   - **Functions**:
>     - Add `fromJsonToLocalTime` in `datetime_utils.dart` to parse and convert time to local.
>   - **Presentation**:
>     - Remove `toLocal()` call in `SksChartBottomTitles` in `sks_chart_labels.dart`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for 37d31c97c328c354bfcdc646d1738412168cff3b. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/sks/sks_chart/data/models/sks_chart_data.dart
+++ b/lib/features/sks/sks_chart/data/models/sks_chart_data.dart
@@ -18,10 +18,6 @@ abstract class SksChartData with _$SksChartData {
   factory SksChartData.fromJson(Map<String, dynamic> json) => _$SksChartDataFromJson(json);
 }
 
-// DateTime toLocalPolish(String dateTimeString) {
-//   return DateTime.parse(dateTimeString).toUtc().toLocal();
-// }
-
 extension SksChartDataIListX on IList<SksChartData> {
   double get maxNumberOfUsers {
     return map(
--- a/lib/features/sks/sks_chart/data/models/sks_chart_data.dart
+++ b/lib/features/sks/sks_chart/data/models/sks_chart_data.dart
@@ -1,4 +1,5 @@
 import "package:fast_immutable_collections/fast_immutable_collections.dart";
+import "package:flutter/foundation.dart";
 import "package:freezed_annotation/freezed_annotation.dart";
 
 part "sks_chart_data.freezed.dart";
@@ -9,10 +10,14 @@ abstract class SksChartData with _$SksChartData {
   const factory SksChartData({
     required int activeUsers,
     required int movingAverage21,
-    required DateTime externalTimestamp,
+    @JsonKey(fromJson: toLocalPolish) required DateTime externalTimestamp,
   }) = _SksChartData;
 
   factory SksChartData.fromJson(Map<String, dynamic> json) => _$SksChartDataFromJson(json);
+}
+
+DateTime toLocalPolish(String dateTimeString) {
+  return DateTime.parse(dateTimeString).toUtc().toLocal();
 }
 
 extension SksChartDataIListX on IList<SksChartData> {
--- a/lib/features/sks/sks_chart/data/models/sks_chart_data.dart
+++ b/lib/features/sks/sks_chart/data/models/sks_chart_data.dart
@@ -2,6 +2,8 @@ import "package:fast_immutable_collections/fast_immutable_collections.dart";
 import "package:flutter/foundation.dart";
 import "package:freezed_annotation/freezed_annotation.dart";
 
+import "../../../../../utils/datetime_utils.dart";
+
 part "sks_chart_data.freezed.dart";
 part "sks_chart_data.g.dart";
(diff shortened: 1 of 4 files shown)


mikolaj-jalocha:
I've one afterthought.

I think I wrongly formulated the task and current situation is not perfect in terms of UX. 

1. Everywhere else we use our polish timezone to show hours: parkings, opening hours etc. SKS'd be an exception - not good. 
2. **I think we should convert server's time (which is not polish I guess, or it's only winter or summertime) to polish timezone taking into account winter/summertime** to be consistent with other parts of the code. Sorry for the confusion, it was my mistake because I wrongly formulated the task. 
3. As I can see from the pic you launched, chart on the right has much less hours shown.

24bartixx:
1) Tbh I didn't like the idea of converting it to local time as well. Why?
* it's very unlikely that a user will check SKS activity chart
* even if user for some reason does it he/she would rather want the time in Polish zone
2) I am not 100% sure but I think it should work in both winter and summer time as I convert it to UTC and the it'd be reasonable to assume that API time takes it into consideration
3) The hours range and the interval between them is actually the same on the pic. The only difference is the time zone. However, I've noticed that the shown hours on the chart can be dependent on the platform.

The question now is what to do with it? Delete the PR and forget? Play with summer and wintertime? (but it might be already accurate idk) @mikolaj-jalocha

mikolaj-jalocha:
> 1. Tbh I didn't like the idea of converting it to local time as well. Why?
> 
> * it's very unlikely that a user will check SKS activity chart
> * even if user for some reason does it he/she would rather want the time in Polish zone
> 
> 2. I am not 100% sure but I think it should work in both winter and summer time as I convert it to UTC and the it'd be reasonable to assume that API time takes it into consideration
> 3. The hours range and the interval between them is actually the same on the pic. The only difference is the time zone. However, I've noticed that the shown hours on the chart can be dependent on the platform.
> 
> The question now is what to do with it? Delete the PR and forget? Play with summer and wintertime? (but it might be already accurate idk) @mikolaj-jalocha

We can't assume that user won't check SKS chart :-) In such case its implementation'd be pointless. 
If we stay with user's local timezone, everywhere else  eg. parkings should be converted as well. 

Let's stick with this PR and play with summer and winter time. Sorry for the fuss.

24bartixx:
Is the given time in API a summer or winter time? Do we know which one it is? Isn't it already correctly transformed on the backend?

mikolaj-jalocha:
> Is the given time in API a summer or winter time? Do we know which one it is? Isn't it already correctly transformed on the backend?

I guess it's not. That's the task objective. Better check on emulator (be sure to set location to PL)

24bartixx:
@mikolaj-jalocha 
Do I understand correctly this assignment? I have to convert time if the user has another date on the phone like 16th July for example? I changed time on my phone and nothing happens on the chart. Should I decrement the hours? Is that the expected behavior. 

We don't really have an option to check if time is correctly passed in API until the time change

mikolaj-jalocha:
> @mikolaj-jalocha Do I understand correctly this assignment? I have to convert time if the user has another date on the phone like 16th July for example? I changed time on my phone and nothing happens on the chart. Should I decrement the hours? Is that the expected behavior.
> 
> We don't really have an option to check if time is correctly passed in API until the time change

Hours displayed at the chart should be "following?" polish timezone (including winter and summertime), whether the user is at Hawaii or in Japan, polish timezone should be displayed at the chart. Now, without your implementation, it's one hour late, in Poland we have 17, and chart is showing 16.

24bartixx:
As I've played with the chart I believe the issue is different. The time is not one hour late but for some reason, the chart doesn't
display the last hour XD Look

image
image

mikolaj-jalocha:
> As I've played with the chart I believe the issue is different. The time is not one hour late but for some reason, the chart doesn't display the last hour XD Look
> 
> image image

yep, it could be. Please check what's going on then. From my observations it apperas that chart? probably cuts some data

24bartixx:
@mikolaj-jalocha @simon-the-shark @tomasz-trela PR done

mikolaj-jalocha (approved):
Good job! The problem was in api, right?

PR #654, opened 2025-03-17
bug(digital-guide): fix accessibility panel colors

thesun901:
Changed accessibility profile card design to sth like this. Colorful background is sadly unachievable, so i decided to put it as black on white ;) 

image


simon-the-shark (approved):
lgtm

PR #681, opened 2025-04-05
refactor: change from watch to read in parkings

tomasz-trela:
While I was doing onboarding I had seen this.

----

> [!IMPORTANT]
> Refactor `FavouriteParkingWidget` to use `ref.read` instead of `ref.watch` for toggling favorite status to avoid unnecessary rebuilds.
> 
>   - **Refactor**:
>     - In `FavouriteParkingWidget`, change from `ref.watch` to `ref.read` for `localFavParkingsRepositoryProvider(parking.id).notifier` in `onPressed` callback to avoid unnecessary rebuilds.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for 4ce92f45a5d6f85e6ffd9aa80e46da8318cd0b34. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/parkings/parkings_view/widgets/parking_favourite.dart
+++ b/lib/features/parkings/parkings_view/widgets/parking_favourite.dart
@@ -15,7 +15,9 @@ class FavouriteParkingWidget extends ConsumerWidget {
 
     return IconButton(
       visualDensity: VisualDensity.compact,
-      onPressed: ref.read(localFavParkingsRepositoryProvider(parking.id).notifier).toggle,
+      onPressed: () async {
+        await ref.read(localFavParkingsRepositoryProvider(parking.id).notifier).toggle();
+      },
       icon:
           isFavorite == null
               ? FavouriteIcon(icon: Icons.error, color: context.colorTheme.whiteSoap)
--- a/lib/features/parkings/parkings_view/widgets/parking_favourite.dart
+++ b/lib/features/parkings/parkings_view/widgets/parking_favourite.dart
@@ -12,7 +12,7 @@ class FavouriteParkingWidget extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final isFavorite = ref.watch(localFavParkingsRepositoryProvider(parking.id));
-    
+
     return IconButton(
       visualDensity: VisualDensity.compact,
       onPressed: ref.read(localFavParkingsRepositoryProvider(parking.id).notifier).toggle,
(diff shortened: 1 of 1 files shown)


lib/features/parkings/parkings_view/widgets/parking_favourite.dart
```diff
@@ -11,7 +11,7 @@ class FavouriteParkingWidget extends ConsumerWidget {
   final Parking parking;
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final favController = ref.watch(localFavParkingsRepositoryProvider(parking.id).notifier);
+    final favController = ref.read(localFavParkingsRepositoryProvider(parking.id).notifier);
```
simon-the-shark, line 14: so afaik your change should be a safe here (because we listen to the notifier's state in the same place too), but generally I would restrain from calling `ref.read` inside of any `build` method, cause we might end up with an old notifier:
Zrzut ekranu 2025-04-5 o 18 11 11

in this case it should be safe, but you can consider small change that will always be safe

if you want to do `ref.read` it's safer to do this inside `onPressed` specifically:
```dart
onPressed: () {
    ref.read(localFavParkingsRepositoryProvider(parking.id).notifier).toggle();
}
```


or do a separate method if you want to separate it from the ui:
```dart
void toggle(WidgetRef ref){
   ref.read(localFavParkingsRepositoryProvider(parking.id).notifier);
}

@override
// ...
onPressed: () => toggle(ref),
// ....
```

simon-the-shark (dismissed):
so here it should be safe, but I would consider changing this to some other option. Also I'm not a that advanced expert on the topic so you do what you prefer, cause in this case it works for sure, but afaik this might be problematic in some cases

lib/features/parkings/parkings_view/widgets/parking_favourite.dart
tomasz-trela: Good catch

lib/features/parkings/parkings_view/widgets/parking_favourite.dart
```diff
@@ -11,11 +11,11 @@ class FavouriteParkingWidget extends ConsumerWidget {
   final Parking parking;
   @override
   Widget build(BuildContext context, WidgetRef ref) {
-    final favController = ref.watch(localFavParkingsRepositoryProvider(parking.id).notifier);
     final isFavorite = ref.watch(localFavParkingsRepositoryProvider(parking.id));
+
     return IconButton(
       visualDensity: VisualDensity.compact,
-      onPressed: favController.toggle,
+      onPressed: ref.read(localFavParkingsRepositoryProvider(parking.id).notifier).toggle,
```
simon-the-shark, line 18:
```suggestion
      onPressed: (){
        ref.read(localFavParkingsRepositoryProvider(parking.id).notifier).toggle
     },
```

it doesn't work the same Tomuś :)

PR #935, opened 2025-11-15
fix: make links accessible in expansion tiles

simplyNoOne:
This PR takes care of two things:
- Children in expansion tile are traversable by keyboard - before, clicking tab would move onto the next expansion header skipping the content altogether - this is most visible in the SKS menu.

https://github.com/user-attachments/assets/d6873da8-8968-43e9-9c72-0a33fd844a06

- Links in HTML widget are visible to screen readers. (I did not manage to make them traversable by keyboard, but they can be seen using touch gestures)

https://github.com/user-attachments/assets/43262fbb-c709-4cf1-b914-7001852d6f01

https://github.com/user-attachments/assets/c38191c6-8e17-4ee6-b016-ff03b45f8608

https://github.com/user-attachments/assets/76c8007c-4fd6-4dfe-b1c1-7c682f6d1131

More explanation about the second issue, because it turned out that the inaccessible links referenced in the issue #802 were NOT a result of using expansion tiles, but rather a problem with HtmlWidget not supporting semantics out of the box.
Upon trial and error I managed to make links readable while the other content is not traversable by gestures, but now, some large links are displayed rather ugly (this doesn't happen often, but it does happen):
image
image

I'd like to know if this end result is acceptable, or if I should keep digging in attempt to add keyboard traversal for links and fix styling issues

 

----

> [!IMPORTANT]
> Make children in expansion tiles keyboard traversable and links in HTML widgets visible to screen readers.
> 
>   - **Behavior**:
>     - Children in expansion tiles are now keyboard traversable (`faq_expansion_tile.dart`, `sks_menu_tiles.dart`).
>     - Links in `MyHtmlWidget` are visible to screen readers but not yet keyboard traversable (`my_html_widget.dart`).
>   - **Extensions**:
>     - Renames `CustomHtmlStylesX` to `CustomHtmlX` and updates `customStylesBuilder` to handle link and text alignment styles (`html_util_extensions.dart`).
>   - **Dependencies**:
>     - Updates `flutter_widget_from_html_core` to a custom git version for accessibility improvements (`pubspec.yaml`).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for e085fa9f240ca156104f49db8bcfcbb115d20f9c. You can [customize](https://app.ellipsis.dev/Solvro/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/widgets/my_html_widget.dart
+++ b/lib/widgets/my_html_widget.dart
@@ -1,10 +1,10 @@
 import "package:flutter/material.dart";
 import "package:flutter_riverpod/flutter_riverpod.dart";
 import "package:flutter_widget_from_html_core/flutter_widget_from_html_core.dart";
-
 import "../features/about_us_view/utils/html_util_extensions.dart";
 import "../theme/app_theme.dart";
 import "../utils/html_to_plain_text.dart";
+import "../utils/launch_url_util.dart";
 
 /// Widget that renders HTML. Supports semantics for screen readers and custom ToPWR styling.
 
@@ -20,6 +20,8 @@ class MyHtmlWidget extends ConsumerWidget {
         html,
         textStyle: textStyle ?? context.aboutUsTheme.body,
         customStylesBuilder: context.customStylesBuilder,
+        onTapUrl: ref.launch,
+        excludePlainTextFromSemantics: true,
       ),
     );
   }


lib/widgets/my_html_widget.dart
```diff
@@ -17,13 +17,13 @@ class MyHtmlWidget extends ConsumerWidget {
   Widget build(BuildContext context, WidgetRef ref) {
     return Semantics(
```
simon-the-shark: did you try and test how this work with this semantics removed as well? Just a wild guess

simplyNoOne: yes, I did try and it must stay here. This semantics widget is necessary for the reader to correctly read the entire block of test with the links, otherwise it just skips the link text completely while reading.

lib/features/about_us_view/utils/html_util_extensions.dart
```diff
@@ -31,4 +33,38 @@ extension CustomHtmlStylesX on BuildContext {
...
+            if (href.isNotEmpty) {
+              await ref.launch(href);
+            }
+          },
+          child: GestureDetector(
+            onTap: () async {
+              if (href.isNotEmpty) {
+                await ref.launch(href);
+              }
+            },
```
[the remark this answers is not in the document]
simplyNoOne: Unfortunately, this doesn't work either. I think the reason the dots are left behind (or any small element really) is the way the html widget is built. First, it processes the text with the link and lays it out, and then it processes the link itself, tries to put it in as an inline widget, but when it takes more that one line, it gives up and moves the remaining text to a new line. 
TLDR; not the GestureDetector's fault

simplyNoOne:
Good news, achieved the desired effect (SR only sees the links but not plain text) - the entire html widget is read out just as it used to be, but now the user can also hear and open links. The visual layout is back to rights.

Changes made to the package are here:  [forked package](https://github.com/simplyNoOne/flutter_widget_from_html/tree/accessibility)

image
image


https://github.com/user-attachments/assets/f11c8ca9-fb0f-4b80-b1ba-fa017c8b4cdd

@simon-the-shark 
@odominiak

simon-the-shark (approved):
Outstanding job Ania!

PR #946, opened 2025-11-21
fix(calendar): empty state and headerHeight

MarmoPL:
- Added already existing empty state from science circles view, but with changed message.
- Fixed headerHeight value that was preventing the calendar from functioning.


simon-the-shark (approved):
LGTM!

PR #973, opened 2026-01-01
refactor: optimize students club filters bottom sheet anim

simplyNoOne:
The issue is that getting values from providers happens when the bottom sheet opening animation is displayed, and all that work on a single thread causes the anim lag.

In each of the wraps: tags, departments, types, I pulled the consumer creation outside of the loop, so that the controller and selected departments are only read once and not 10 or 20 times.
I removed the theme wrapping the filter chip, I cannot see what difference the theme made.

I cannot say that these changes visibly improved the animation, it still lags, but now is perhaps slightly less suboptimal than in was previously? IDK if there is a point in trying to improve things further, I doubt many people will notice this lag.

----

> [!IMPORTANT]
> Optimized filters bottom sheet animation by reducing redundant provider reads and simplifying widget tree.
> 
>   - **Optimization**:
>     - In `DepartmentsWrap`, `TagsWrap`, and `TypesWrap`, moved provider reads outside loops to reduce redundant reads.
>     - Removed `Theme` wrapping in `MyFilterChip` to simplify widget tree.
>   - **Behavior**:
>     - Added early return in `departmentFiltersFiltered()` in `filters_search_controller.dart` if query is empty.
>   - **Misc**:
>     - Removed unused imports in `departments_wrap.dart`, `tags_wrap.dart`, and `types_wrap.dart`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for 5ae8641e462d08bb6932b00610b9d593bf9bcd65. You can [customize](https://app.ellipsis.dev/Solvro/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/science_club/science_clubs_filters/filters_search_controller.dart
+++ b/lib/features/science_club/science_clubs_filters/filters_search_controller.dart
@@ -40,7 +40,7 @@ IList<ScienceClubType> typeFiltersFiltered(Ref ref) {
 @Riverpod(dependencies: [SearchFiltersController])
 Future<IList<Department>> departmentFiltersFiltered(Ref ref) async {
   final query = ref.watch(searchFiltersControllerProvider);
-  final depts = ref.watch(departmentsRepositoryProvider).requireValue;
+  final depts = await ref.watch(departmentsRepositoryProvider.future);
 
   if (query.isEmpty) return depts.toIList();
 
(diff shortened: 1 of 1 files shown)


lib/features/science_club/science_clubs_filters/filters_search_controller.dart
```diff
@@ -52,7 +55,7 @@ Future<IList<Department>> departmentFiltersFiltered(Ref ref) async {
 @Riverpod(dependencies: [SearchFiltersController])
 Future<IList<Tag>> tagFiltersFiltered(Ref ref) async {
   final query = ref.watch(searchFiltersControllerProvider);
-  final tags = await ref.watch(tagsRepositoryProvider.future);
+  final tags = ref.watch(tagsRepositoryProvider).requireValue;
```
simon-the-shark: why this change? it becomes a racing condition this way, so wouldn't recommend it

simon-the-shark (commented):
i have one reservation, unless this is a big performance game changer I wouldn't do that. except for that all looks okey, safe to merge

lib/features/science_club/science_clubs_filters/filters_search_controller.dart
simplyNoOne: i believe it was marginally faster after this change, but nothing really noticeable

PR #976, opened 2026-01-06
fix: prevent scrolling bug in digital guide

24bartixx:
The issue was that the accesibility widget with cards was had circular progress indicator when it was loading data. Then the indicator was changing to SizedBox.shink()

1) height > 0
2) height = 0 or height >>> 0  
3) weird jump

I implemented cache to store previous accesibility that is fetched asynchrounously causing this loading.
Moreover, I refactor the accesibiltiy list to load the card widgets lazily and used column instead of ListView (it's already inside CustomScrollView and there is not that many cards to render them dynamically imo) for perf.

----

> [!IMPORTANT]
> Implements caching and refactors accessibility list to prevent scrolling bug and improve performance in `active_accessibility_modes_repository.dart`.
> 
>   - **Behavior**:
>     - Implements caching in `ActiveAccessibilityModesRepository` to store previous accessibility data, preventing unnecessary re-fetching.
>     - Refactors accessibility list to load card widgets lazily using `Column` instead of `ListView` for performance improvement.
>   - **Repository**:
>     - Changes `activeAccessibilityModesRepository` to a class `ActiveAccessibilityModesRepository` with caching logic in `active_accessibility_modes_repository.dart`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for 77dd3d6c0e815985ae7b28e8576770eb948fd6ca. You can [customize](https://app.ellipsis.dev/Solvro/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/digital_guide/tabs/accessibility_dialog/data/active_accessibility_modes_repository.dart
+++ b/lib/features/digital_guide/tabs/accessibility_dialog/data/active_accessibility_modes_repository.dart
@@ -1,3 +1,4 @@
+import "dart:async";
 import "package:fast_immutable_collections/fast_immutable_collections.dart";
 import "package:riverpod_annotation/riverpod_annotation.dart";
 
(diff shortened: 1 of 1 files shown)


lib/features/digital_guide/presentation/widgets/accessibility_information_cards_list.dart
```diff
@@ -47,108 +50,117 @@ class AccessibilityInformationCardsList extends ConsumerWidget {
...
+  }
 
-        final widgets = currentModes.map((mode) => cards[mode]).toList();
+  Map<ModeWithKey, AccessibilityInformationCard> _getCardsMap(
+    BuildContext context, {
+    required AccessibilityInformationLevelsInput levels,
+    required AccessibilityInformationOptionalCommentsInput? comments,
+    required String prefix,
+    required String Function(String) accessibilityLevelType,
+  }) {
```
simon-the-shark: instead of creating a method that returns a map I would just perfer to have a declarative-style switch (even inline in the build)

simon-the-shark (approved):

PR #1021, opened 2026-02-08
fix(sks-favs): offline view

simplyNoOne:
Offline view didn't show up because subscriptions fetch was not using custom safe get.

Decided to use getAndCacheDataWithTranslation function to ensure translations and keep implementation consistent with pretty much every other repository in the project. However, because subscriptions of the user can change whenever, cache should always be invalidated.

https://github.com/user-attachments/assets/3ccd3175-2fa1-4cdc-9ff5-a38ff38c460b

----

> [!IMPORTANT]
> Use `getAndCacheDataWithTranslation` for subscription data in `SksFavouriteDishesRepository`, ensuring translations and always invalidating cache.
> 
>   - **Behavior**:
>     - Use `getAndCacheDataWithTranslation` for fetching subscription data in `SksFavouriteDishesRepository` to ensure translations and consistency.
>     - Cache for subscription data is always invalidated to handle dynamic changes.
>   - **Misc**:
>     - Removed use of custom safe get for fetching subscriptions.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for f31746e2923796baad3d21a1eba3689dd23ebd49. You can [customize](https://app.ellipsis.dev/Solvro/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/sks/sks_favourite_dishes/data/repository/sks_favourite_dishes_repository.dart
+++ b/lib/features/sks/sks_favourite_dishes/data/repository/sks_favourite_dishes_repository.dart
@@ -38,9 +38,14 @@ class SksFavouriteDishesRepository extends _$SksFavouriteDishesRepository {
           .castAsObject,
       if (deviceKey != null)
         ref
-            .read(restClientProvider)
-            .get<Map<String, dynamic>>(_api + _subscriptionsEndpoint + deviceKey)
-            .then((val) => SksFavouriteDishesResponse.fromJson(val.data!)),
+            .getAndCacheDataWithTranslation(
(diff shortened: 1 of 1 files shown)


simon-the-shark (approved):
yeah this looks cool. One think is that the subscriptions can't change ANYTIME, they can only change when we perform a mutation on our side - but sadly our cache system does not really support manual on-demand cache invalidation mutation (this is basically only remote mutation in the whole app, so we didn't have a reason for that) - that's why this is fully acceptable I think. Mergin!

PR #1028, opened 2026-02-13
fix(sci-clubs): overlapping issue

24bartixx:
# How it was: 

image

# Issue
The search icon (part of expandable widget on click) is positioned in a stack so it overlaps the text. However, the text was part of a full-width widget that was used in the home screen 
image
image
Therefore, it would be difficult to position it exactly after the text every time (which could result in overflow, etc.) Besides looking at bottom sheets and filter designs on Dribble and Google I realized that the expandable search icon next to section title is not a common way to go. It would be the best to position it below the header but then the expansion would be lost (and the hard work of the person who did it)
image

# Solution
Eventually, I moved the Clear button to the bottom next to Submit button and moved the search icon to the right. Now, on expansion the search bar has the same width as it used to and the expansion animation looks better, as it expands towards the left only now.

# Screenshots
image
image
image

# Inspiration with the buttons
image
image

What do you think about this refactor? If you have better idea let me know

----

> [!IMPORTANT]
> Fix overlapping issue in science club filters by repositioning UI elements and refactoring buttons.
> 
>   - **UI Adjustments**:
>     - Moved Clear button next to Submit button in `filters_sheet.dart`.
>     - Repositioned search icon to the right in `filters_search.dart`.
>     - Adjusted `FiltersHeader` in `filters_header.dart` for better alignment.
>   - **Widget Refactoring**:
>     - Renamed `ApplyFiltersButton` to `FiltersButton` and refactored to handle both clear and apply actions in `filters_button.dart`.
>   - **Configuration Changes**:
>     - Updated `searchFilterPadding` to 20.0 and `radius` to 16 in `ui_config.dart`.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for e0914f37ff921a1b2023d005f8e7ae1dd01cdb74. You can [customize](https://app.ellipsis.dev/Solvro/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/science_club/science_clubs_filters/filters_controller.dart
+++ b/lib/features/science_club/science_clubs_filters/filters_controller.dart
@@ -56,6 +56,7 @@ class SelectedTypeController extends _$SelectedTypeController with FilterControl
 }
 
 @Riverpod(
+  keepAlive: true,
   dependencies: [SelectedDepartmentController, SelectedTagController, SelectedTypeController, SearchFiltersController],
 )
 bool areFiltersEnabled(Ref ref) {
--- a/lib/features/science_club/science_clubs_filters/filters_sheet.dart
+++ b/lib/features/science_club/science_clubs_filters/filters_sheet.dart
@@ -24,6 +24,7 @@ class FiltersSheet extends ConsumerWidget {
   @override
   Widget build(BuildContext context, WidgetRef ref) {
     final sheetHeight = useFiltersSheetHeight(context);
+
     return ProviderScope(
       overrides: [areFiltersEnabledProvider],
       child: Semantics(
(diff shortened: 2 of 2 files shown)


odominiak:
As for the appearance, I really like it.

I have only one question about the behavior of the "Clear" button. Previously, the button was inactive (grey) until you clicked on some filter, and it only became active after the mentioned click.
Whether this behavior has been preserved here?

I’m also wondering whether it would be possible for the “Clear” button to appear only after selecting a filter, and otherwise show only the “Apply” button?
If for whatever reason that’s not possible, then please just make sure the “Clear” button remains greyed out when no filter is selected.

24bartixx:
Done! Now it hides when no filters

simon-the-shark (approved):

PR #1041, opened 2026-02-23
fix: countdown text overflow

24bartixx:
image

I restricted the number of lines to 2, as I had idea how to display the whole text on screens with narrow resolutions. Do you have another idea to approach the issue @odominiak ?

> [!IMPORTANT]
> Add `maxLines: 2` to `Text` widget in `ExamSessionCountdown` to prevent text overflow.
> 
>   - **UI Fix**:
>     - Add `maxLines: 2` to `Text` widget in `ExamSessionCountdown` to prevent text overflow.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=Solvro%2Fmobile-topwr&utm_source=github&utm_medium=referral)<sup> for 5f1508adbac4bd35bf21f883c73c2e06861dc173. You can [customize](https://app.ellipsis.dev/Solvro/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/features/academic_calendar/widgets/countdown_widget/exam_session_countdown.dart
+++ b/lib/features/academic_calendar/widgets/countdown_widget/exam_session_countdown.dart
@@ -69,6 +69,7 @@ class ExamSessionCountdown extends ConsumerWidget {
                         academicCalendar.localizeDaysCounterMessage(context),
                         style: context.textTheme.bodyLarge?.copyWith(color: context.colorScheme.surface),
                         textScaler: const TextScaler.linear(0.97),
+                        maxLines: 2,
                       ),
                     ),
                   ],


24bartixx:
> Please show extra large fonts on screenshot 👉 👈 @24bartixx

image

Basically, the text for large fonts is not visible because the height of the widget is restricted. 
I could move the text to the bottom below the numbers if that's okay @odominiak  Maybe it'd would solve the layout problems.

Additionally, what do you think about the text and these numbers scaling Szymon mentioned above?

Which option would be better?

simon-the-shark (approved):

odominiak:
If u have an good idea for it, then go for it, but if not lets leave it like this

PR #1043, opened 2026-02-27
feat: auto refresh of recently played

MarmoPL:
Refreshes every 60 seconds


simon-the-shark (commented):
Let's maybe try 20s not 60s - but otherwise good to go

PR #1126, opened 2026-03-28
feat(sks-menu): new sks opening hours model

### LINKED ISSUE #1034: feat(sks-menu): integrate with new sks opening hours model
https://api.topwr.solvro.pl/api/v1/sks_opening_hours

Image

MarmoPL:
Screenshot_20260328_191930


lib/features/sks/sks_menu/data/models/sks_opening_hours.dart
```diff
@@ -7,19 +7,19 @@ part "sks_opening_hours.g.dart";
...
-  factory SksOpeningHours.fromJson(Map<String, dynamic> json) => _$SksOpeningHoursFromJson(json);
+  factory SksOpeningHours.fromJson(Map<String, dynamic> json) {
+    final list = json["data"] as List<dynamic>;
+    if (list.isEmpty) {
+      throw const FormatException("sks_opening_hours: 'data' array is empty");
+    }
+    final data = list.first as Map<String, dynamic>;
+    return SksOpeningHours(openingHours: OpeningHours.fromJson(data));
+  }
 }
```
simon-the-shark: definitely not how we wanna do it.

1st of all, no reason to override the freezed constructor pretty much never 

2nd of all, there should be a field `language` and we should pick `pl` or `eng` version based on app's current language. (best effort, so with fallback to the other one)

simon-the-shark (changes requested):
@odominiak can u insert some entry for the other language as well for testing so Marcin can adjust this and test

MarmoPL:
I think I've done it. Currently the endpoint returns empty list, so I haven't tested it with data yet. But I've made sure, that if the data is not available, the whole menu will not give an error, just times would not get displayed at all.

simon-the-shark (approved):

PR #1152, opened 2026-05-11
fix(calendar): remove slashes from events location

### LINKED ISSUE #1148: fix(calendar): remove slashes from events location
Image

mikolaj-r:
These backslashes come from parsing ICS files. My solution only fixes the visible text in the calendar view. If changes in the global scope are preferable, editing the model file or handling it during backend parsing would be a better option.


simon-the-shark (approved):
very nice.

PR #1194, opened 2026-06-07
feat: full Apple CarPlay integration

MarmoPL:
Previously, the Apple CarPlay integration was not actually finished. On iOS, a complete UI has to be defined, unlike on Android Auto.

This PR introduces a full CarPlay interface and lets users start radio playback even when the phone app isn’t open.

It also resolves the duplicated album and artist text shown in the player controls.


simon-the-shark:
integration tests are not passing

lib/features/radio_luz/service/radio_audio_handler.dart
```diff
@@ -67,17 +65,6 @@ class RadioAudioHandlerBridge extends BaseAudioHandler with SeekHandler, Widgets
...
-    WidgetsBinding.instance.addObserver(this);
-  }
-
-  @override
-  void didChangeAppLifecycleState(AppLifecycleState state) {
-    // Stop playback when app is detached (killed)
-    if (state == AppLifecycleState.detached) {
-      unawaited(stop());
-    }
   }
```
simon-the-shark, line 81: why do we drop this?

MarmoPL: image

MarmoPL: But wait, Greptile can be wrong on this...
@greptile what about Android? Is it affected by this change?

MarmoPL: @greptile the app should not survive app death (it should stop playing when closed from recents, on iOS and Android), but it should not stop playing when connected to CarPlay or Android Auto. What's the solution then?

simon-the-shark (approved):

PR #1196, opened 2026-06-10
feat(activity-days): add Timetable view

### LINKED ISSUE #1114: feat(activity-days): add Timetable view
WIP
> backend's issues regarding activity days are linked in parent issue (#1073)

mikolaj-r:
I used a design similar to Calendar view and it behaves similarly to calendar in terms of different font sizes for example. It is designed to handle events across several days. In case if it will be only one day and the date text shouldn't be visible, it can be easily deleted. Also mocked a couple of entries for testing. Lacks search function, not sure if it is needed in this view.

image


simon-the-shark (approved):
sorry for the delay but all look pretty nice.

PR #1218, opened 2026-07-23
fix(sks-menu): single click on menu position shoudn't trigger haptics

### LINKED ISSUE #1215: fix(sks-menu): single click on menu position shoudn't trigger haptics
when clicking in menu view on a menu item it causes haptic response but no action (there should be no action. the issue is that the haptics are wrongly added in such case).

also be carefull, in the sks fav view single tap on teh same widget has an action so the haptics are right there

mikolaj-r:
Haptics are triggered only on action, including a double tap in sks menu (not sure if this is desired).


simon-the-shark (approved):
yeah I believe it is desired, nice job with all these PRs

